PR #24, opened 2018-04-03
Replace logxi with hclog

kalafut:


path_role_test.go
```diff
@@ -17,7 +16,8 @@ func getBackend(t *testing.T) (logical.Backend, logical.Storage) {
 	b := Backend()
 
 	config := &logical.BackendConfig{
-		Logger: logging.NewVaultLogger(log.LevelTrace),
+		Logger: log.New(&log.LoggerOptions{Level: log.Trace}),
```
chrishoffman: I'm not sure the difference, but why aren't we creating a logger with vault/helper/logging?

kalafut: Copy-n-paste momentum is the only reason. Some weren't changed with the helper before, though they probably should be to pick up the JSON detection. I'll fix it.

PR #27, opened 2018-06-01
Fix test failure

tyrannosaurus-becks:
The test failure was because the test expected a field to be returned that wasn't.


path_config.go
```diff
@@ -66,6 +66,7 @@ func (b *kubeAuthBackend) pathConfigRead() framework.OperationFunc {
 					"kubernetes_host":    config.Host,
 					"kubernetes_ca_cert": config.CACert,
 					"pem_keys":           config.PEMKeys,
+					"token_reviewer_jwt": config.TokenReviewerJWT,
```
jefferai: We don't return sensitive information in reads anymore, so this should not be added. The test needs to be updated to not require it, or be removed.

tyrannosaurus-becks: :+1:

PR #28, opened 2018-06-01
Add bound cidrs for roles

tyrannosaurus-becks:
Closes https://github.com/hashicorp/vault/issues/4662

### DIFF
--- a/path_login.go
+++ b/path_login.go
@@ -328,16 +328,10 @@ func (b *kubeAuthBackend) pathLoginRenew() framework.OperationFunc {
 			return nil, fmt.Errorf("role %s does not exist during renewal", roleName)
 		}
 
-		// Check for a CIDR match.
-		if !cidrutil.RemoteAddrIsOk(req.Connection.RemoteAddr, role.BoundCIDRs) {
-			return nil, errors.New("renewal request originated from invalid CIDR")
-		}
-
 		resp := &logical.Response{Auth: req.Auth}
 		resp.Auth.TTL = role.TTL
 		resp.Auth.MaxTTL = role.MaxTTL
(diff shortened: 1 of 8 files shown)


path_login.go
```diff
@@ -77,6 +78,11 @@ func (b *kubeAuthBackend) pathLogin() framework.OperationFunc {
 			return logical.ErrorResponse(fmt.Sprintf("invalid role name \"%s\"", roleName)), nil
 		}
 
+		// Check for a CIDR match.
+		if !cidrutil.RemoteAddrIsOk(req.Connection.RemoteAddr, role.BoundCIDRs) {
+			return nil, errors.New("renewal request originated from invalid CIDR")
```
jefferai: login request, not renewal

path_login_test.go
```diff
@@ -53,10 +53,11 @@ func setupBackend(t *testing.T, noPEMs bool) (logical.Backend, logical.Storage)
...
 	req = &logical.Request{
-		Operation: logical.CreateOperation,
-		Path:      "role/plugin-test",
-		Storage:   storage,
-		Data:      data,
+		Operation:  logical.CreateOperation,
+		Path:       "role/plugin-test",
+		Storage:    storage,
+		Data:       data,
+		Connection: &logical.Connection{},
```
jefferai: Rather than adding empty Connection blocks in tests, you're better off checking for nil connection info in the actual login/renewal code, and failing if cidrs are defined but it's nil.

tyrannosaurus-becks: Hm, good point. It's way less code too.

idcmp:
Amazing! Thank you! :)

Issue #35, opened 2018-07-20
Unable to login to auth/kubernetes backend using JWT


falzm:
Hi

I'm trying to get the Vault Kubernetes auth working but I'm stumbling on a weird problem when trying to login to Vault using a k8s pod's service account (I'm running Kubernetes on GKE):

```
### In the pod container

$ cat /var/run/secrets/kubernetes.io/serviceaccount/token ; echo
eyJhbGciOiJSUzI1NiIsImtpZCI6IiJ9.eyJpc3MiOiJrdWJlcm5ldGVzL3NlcnZpY2VhY2NvdW50Iiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9uYW1lc3BhY2UiOiJhIiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9zZWNyZXQubmFtZSI6ImFwcC1hLXRva2VuLXR6ZjdiIiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9zZXJ2aWNlLWFjY291bnQubmFtZSI6ImFwcC1hIiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9zZXJ2aWNlLWFjY291bnQudWlkIjoiZjA4MDlhYmQtOGMyNy0xMWU4LTkwNjUtNDIwMTBhODQwMDE2Iiwic3ViIjoic3lzdGVtOnNlcnZpY2VhY2NvdW50OmE6YXBwLWEifQ.JASwQlSOZmqQfB_EKw__XB4Vsogl3jlQHDzKWBGSuWhahgdhC5v3i87k0463os1GA6pCY6DG5eFayYq0l-B-iXv2fvUkF-hSjfeTNauVJNu4PFRlxv9TahPznGYHrpxQ4PodmbilumkEXE9R1VcQEJ9IEBopyMqiudnw90wQqNt-WHNWlPaS9I8pZMLWa89ksQ_y3XeySToUfz3w89LloUM1cBDOnySNaZQR_se4ugc1aqWo0zvnA6ydSamNF0NmYbjLq1u5pCJx_eAjtRECo-UOfaiYrWn-FOEXjX_w1ojc7xobLJ0DNzzVv665ZHEgcvaewxoFWB4ctkDubtTbuQ

[...]

PR #37, opened 2018-07-23
Allow trailing newline char in backend config JWT

falzm:
Fixes https://github.com/hashicorp/vault-plugin-auth-kubernetes/issues/35


jefferai:
Can you make this a `strings.TrimSpace` instead?

falzm:
Done.

PR #40, opened 2018-08-05
Add MIME type headers to the TokenReview request

rallyben:
Observed this headers were missing while experimenting with a Flask
application that mimicked the Kubernetes TokenReview endpoint.


rallyben:
I totally agree this commit should not be accepted on the basis of my Flask application code.  I mentioned that only because Flask is particular about reading the body of requests.

The headers in question are recommended here in the Kubernetes API usage guide: https://kubernetes.io/docs/reference/using-api/api-concepts/#alternate-representations-of-resources

PR #43, opened 2018-10-19
Support new Projected Service Account tokens

catsby:
Initial work to support new [Projected Service Account Tokens][1], which are in beta in Kubernetes v1.12.0. The dependencies are messed up because I had to revert 30aa9f5 in order to get the plugin to compile. 

There are no new additional bounds or restrictions added, other than the new tokens come with an expiration which is automatically verified by the `jwt` library. Future work will include bounding by pod name, audiences, etc. 

Considered a work-in-progress mostly due to dependencies, but would like to get a review started for the work on `path_login.go` and `path_login_test.go`.

[1]: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection

The formats of the Service Account tokens differ significatnly. The "traditional" / default token has this format:

```json
{
  "iss": "kubernetes/serviceaccount",
  "kubernetes.io/serviceaccount/namespace": "default",
  "kubernetes.io/serviceaccount/secret.name": "default-token-pkdn8",
  "kubernetes.io/serviceaccount/service-account.name": "default",
  "kubernetes.io/serviceaccount/service-account.uid": "b389b3b2-d302-11e8-b14c-080027e55ea8",
  "sub": "system:serviceaccount:default:default"
}
```

Where the new projected tokens look like this:

```json
{
  "aud": [
    ""
  ],
  "exp": 1539638092,
  "iat": 1539630892,
  "iss": "api",
  "kubernetes.io": {
    "namespace": "default",
    "pod": {
      "name": "vault",
      "uid": "97da248d-d0ae-11e8-accb-0800274ada6f"
    },
    "serviceaccount": {
      "name": "default",
      "uid": "8b15a81a-d09b-11e8-9c49-0800274ada6f"
    }
  },
  "nbf": 1539630892,
  "sub": "system:serviceaccount:default:default"
}
```

------

Remaining:

- [x] Dependencies


path_login.go
```diff
@@ -280,10 +280,57 @@ func (b *kubeAuthBackend) parseAndValidateJWT(jwtStr string, role *roleStorageEn
...
+	// at the top level
+	Kubernetes *projectedServiceToken `mapstructure:"kubernetes.io"`
+	Expiration int64                  `mapstructure:"exp"`
+	IssuedAt   int64                  `mapstructure:"iat"`
+}
+
+// uid returns the UID for the service account, preferring the projected service
+// account value if found
+func (s *serviceAccount) uid() string {
+	if s.Kubernetes != nil && s.Kubernetes.ServiceAccount != nil {
```
tyrannosaurus-becks: Should there be an additional check on this like `&& s.Kubernetes.ServiceAccount != ""`?

path_login.go
```diff
@@ -280,10 +280,57 @@ func (b *kubeAuthBackend) parseAndValidateJWT(jwtStr string, role *roleStorageEn
...
+		return s.Kubernetes.ServiceAccount.UID
+	}
+	return s.UID
+}
+
+// name returns the name for the service account, preferring the projected
+// service account value if found. This is "default" for projected service
+// accounts
+func (s *serviceAccount) name() string {
+	if s.Kubernetes != nil && s.Kubernetes.ServiceAccount != nil {
```
tyrannosaurus-becks: Should this one have an additional check as well for whether the string itself is populated?

path_login.go
```diff
@@ -280,10 +280,57 @@ func (b *kubeAuthBackend) parseAndValidateJWT(jwtStr string, role *roleStorageEn
...
+		return s.Kubernetes.ServiceAccount.Name
+	}
+	return s.Name
+}
+
+// namespace returns the namespace for the service account, preferring the
+// projected service account value if found
+func (s *serviceAccount) namespace() string {
+	if s.Kubernetes != nil {
+		return s.Kubernetes.Namespace
```
tyrannosaurus-becks: Ditto regarding the additional check question.

path_login.go
catsby: The two formats differ quite a bit (shown in my updated PR comment above). It's either or, so either the top level UID is there, or this sub `Kubernetes` exists, but not both. If the `Kubernetes` struct exits, then the top level `uid` is `""`, so even if `Kubernetes.ServiceAccount.UID` is `""`, the fact that `ServiceAccount` itself is *not* nil means that the top level `UID` is `""` anyway. Does that make sense?

path_login.go
catsby: Same reply as above

path_login.go
tyrannosaurus-becks: Yes! Cool!

PR #50, opened 2018-11-29
updated dependencies

shelby-moore-hs:


catsby:
Hey there! Thanks for doing this. Can you share what command you ran here? Seems all the digests were removed, and an import block was replaced with a digest. I'm not sure if this is a new `dep` format or something, just looking to verify 😄

shelby-moore-hs:
I just ran `dep ensure` - let me know if I should run something else. Is it okay if I update the README to document what needs to be run to grab dependencies?

catsby:
Hey @shelby-moore-hs - I suspect you may have used an older version of `dep`, ensure shouldn't change `Gopkg.lock` (pretty sure..). The changes there are telling too; removal of `digest` and `input-imports`, both added in dep 0.5 I think:

- https://golang.github.io/dep/blog/2018/07/25/announce-v0.5.0.html

To save you more hassle, I went ahead and ran `dep ensure` with `dep 0.5` and merged the results: https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/52

I'm going to close this, thanks! 

For the rest:

- If you can rebase your original PR, that would be awesome, but I'll totally understand if you'd rather me do it since I've already made you jump through enough hoops
- Re: README, it's absolutely okay if you have the time 😄

catsby:
Actually, I'm going to cherry-pick https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/48/commits/4d5cc1961fbf6b01f4ff54b8a0b97b3fd00171c0 into a new PR to save you trouble, thanks!

PR #58, opened 2019-03-12
Glob/Wildcard support for roles

nathkn:
Solves #47 

Allows for roles to specify their namespaces as globbed patterns:
```sh
vault write auth/kubernetes/role/my-role \
    bound_service_account_names=my-globbed-sa-* \
    bound_service_account_namespaces=my-globbed-namespace-* \
    policies=default \
    ttl=1h
```

Two main changes:
1) Use `StrListContainsGlob` instead of `StrListContains` to allow for
   globbed names/namespaces - seemed like the path of least resistance

2) Update the `setupBackend` test function to accept the test role's allowed SA
   name/namespace as parameters, allowing testing of globbed namespaces

Not sure if more logic around this is needed since kubernetes names are pretty limited with regard to special characters anyway.

### DIFF
--- a/path_login.go
+++ b/path_login.go
@@ -193,14 +193,14 @@ func (b *kubeAuthBackend) parseAndValidateJWT(jwtStr string, role *roleStorageEn
 
 			// verify the namespace is allowed
 			if len(role.ServiceAccountNamespaces) > 1 || role.ServiceAccountNamespaces[0] != "*" {
-				if !strutil.StrListContains(role.ServiceAccountNamespaces, sa.namespace()) {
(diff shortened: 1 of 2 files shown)


mterron:
Can someone please take a look at this PR. It'll make k8s auth actually usable.

jefferai:
Thanks!

idcmp:
Hey just catching up on this to track issue #47.

Ordering of StrListContainsGlob is undefined, so an entry for "foo" and "f*" may get the "f*" match, or may get the "foo" wildcard. This makes it impossible to have hardcoded service accounts wild a wildcard catch-all (which is likely the most desirable usage of wildcards).

PR #66, opened 2019-06-28
Remove check for serviceaccount and namespace being *

logandavies181:
Since #48 (#53) we're able to sensibly allow all serviceaccounts in all namespaces to authenticate to Vault, for example mapping them to a templated policy.
Since #58 we're able to get around the check by doing this: 
````
vault write auth/kubernetes/role/myrole \
bound_service_account_names=a*,b*,c*,d*,e*,f*,h*,i*,j*,k*,l*,m*,n*,o*,p*,q*,r*,s*,t*,u*,v*,w*,x*,y*,z*,1*,2*,3*,4*,5*,6*,7*,8*,9*,0* \ 
bound_service_account_namespaces=* policies=my_templated_policy
````
This PR is to remove the now unnecessary check for bound_service_account_name and bound_service_account_namespace being *

### DIFF
--- a/path_login.go
+++ b/path_login.go
@@ -125,7 +125,7 @@ func (b *kubeAuthBackend) pathLogin() framework.OperationFunc {
 					"service_account_name":        serviceAccount.name(),
 					"service_account_namespace":   serviceAccount.namespace(),
 					"service_account_secret_name": serviceAccount.SecretName,
-					"role": roleName,
+					"role":                        roleName,
 				},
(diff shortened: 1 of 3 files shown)


james-atwill-hs:
This is outstanding to finally resolve #47 .. is there a way to move this along?

buckner:
Any chance this could get looked at? @logandavies181

tyrannosaurus-becks:
Closing due to inactivity and to lack of signing the contributor agreement. However, I do still think this is an important request and will either submit a similar PR or will chat with folks about it further on #47.

PR #67, opened 2019-07-02
Tokenutilize

jefferai:


go.mod
```diff
@@ -4,6 +4,7 @@ go 1.12
 
 require (
 	github.com/briankassouf/jose v0.9.2-0.20180619214549-d2569464773f
+	github.com/go-test/deep v1.0.2-0.20181118220953-042da051cf31
```
tyrannosaurus-becks: I think we might need a `go mod vendor` as well.

jefferai: Oh I didn't realize it was vendoring. Will do.

PR #69, opened 2019-08-05
Remove any trailing slash before doing token review

### LINKED ISSUE #59: Configuration fails if Host ends with a slash
If you configure the kubernetes host with a trailing slash (e.g. https://kube.example.com/), all requests will fail with a cryptic error message (it gets a 404 from Kubernetes, which results in a 500 from Vault).

This is caused by https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/master/token_review.go#L76, which adds a slash before the start of the path, regardless of whether the host ends in a slash.

Expected behavior: The host configuration should work either with or without the trailing slash. Whether that happens when configuration is entered or in the token_review.go, I don't care, but I shouldn't have to read your source code because I entered a valid URL as the host.

catsby:
This PR removes any trailing `/` in the configured Kubernetes host information before performing the token review, to ensure we build a valid URL.

Fixes #59

### DIFF
--- a/token_review.go
+++ b/token_review.go
@@ -73,7 +73,7 @@ func (t *tokenReviewAPI) Review(jwt string) (*tokenReviewResult, error) {
 	}
 
 	// Build the request to the token review API
-	url := fmt.Sprintf("%s/apis/authentication.k8s.io/v1/tokenreviews", t.config.Host)
+	url := fmt.Sprintf("%s/apis/authentication.k8s.io/v1/tokenreviews", strings.TrimSuffix(t.config.Host, "/"))
 	req, err := http.NewRequest("POST", url, bytes.NewBuffer(trJSON))
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


token_review.go
```diff
@@ -73,7 +73,7 @@ func (t *tokenReviewAPI) Review(jwt string) (*tokenReviewResult, error) {
 	}
 
 	// Build the request to the token review API
-	url := fmt.Sprintf("%s/apis/authentication.k8s.io/v1/tokenreviews", t.config.Host)
+	url := fmt.Sprintf("%s/apis/authentication.k8s.io/v1/tokenreviews", strings.TrimSuffix(t.config.Host, "/"))
```
chrishoffman: Can you just use https://golang.org/pkg/path/#Join here?

catsby: We could, I briefly looked for some kind of `join` but I was looking in the `http` package. Interesting enough, trying it out in the playground shows `path.Join` removing one of the `/` from the url scheme

- https://play.golang.org/p/DoHiFCf4N9R

That url seems to work in a browser and curl, but it still seems weird, unless I constructed that example incorrectly. 

Is there any other advantage to changing from the existing `fmt.Sprintf` to `path.Join`?

Issue #65, opened 2019-06-27
Better support for K8S projected/scoped tokens


simonswine:
This is a general service account token, which can be used just fine with this plugin. But it also allows access to the kubernetes API itself.

```
{
  "iss": "kubernetes/serviceaccount",
  "kubernetes.io/serviceaccount/namespace": "gitlab-runner",
  "kubernetes.io/serviceaccount/secret.name": "christian-simon--test-token-pml6t",
  "kubernetes.io/serviceaccount/service-account.name": "christian-simon--test",
  "kubernetes.io/serviceaccount/service-account.uid": "f101dff5-98b6-11e9-ac8d-42010a8400cb",
  "sub": "system:serviceaccount:gitlab-runner:christian-simon--test"
}
```

I was trying to setup a projected/scoped token, which can't be used to access k8s API server and comes with a custom audiences. The pod has been created like that:

```
kind: Pod
apiVersion: v1
metadata:
  name: nginx
spec:
  containers:
  - image: nginx
    name: nginx
    volumeMounts:
    - mountPath: /var/run/secrets/tokens
      name: vault-token
  volumes:
  - name: vault-token
    projected:
      sources:
      - serviceAccountToken:
          path: vault-token
          expirationSeconds: 7200
          audience: https://vault.mycorp.net
```

This is the resulting JWT:

```
{
  "aud": [
    "https://vault.mycorp.net"
  ],
  "exp": 1561638806,
  "iat": 1561631606,
  "iss": "https://container.googleapis.com/v1/projects/mycorp-gitlab-workers/locations/europe-west1-b/clusters/mycorp-gitlab-workers",
  "kubernetes.io": {
    "namespace": "gitlab-runner",
    "pod": {
      "name": "runner-9fswzwaf-project-382-concurrent-0l99cc",
      "uid": "ff505f37-98c6-11e9-ac8d-42010a8400cb"
    },
    "serviceaccount": {
      "name": "christian-simon--test",
      "uid": "f101dff5-98b6-11e9-ac8d-42010a8400cb"
    }
  },
  "nbf": 1561631606,
  "sub": "system:serviceaccount:gitlab-runner:christian-simon--test"
}
```

Now trying this token with the auth plugin fails as the issuer is not as expected:

```
Code: 500. Errors:

* claim "iss" is invalid
```

I also can't use the JWT auth plugin, because Google's GKE doesn't give me access to the signing keys.

So I am proposing:

[...]

PR #70, opened 2019-08-06
Add option for specifying the jwt issuer in the configuration.

JWong101:
The issuer for projected service account tokens may not necessarily be "kubernetes/serviceaccount". This pull request provides better support for projected tokens, by allowing the user to specify the issuer during the configuration.


JWong101:
Also, I decided to add the option for specifying the issuer in the configuration rather than the role (which was what #65 was suggesting), since the issuer should be the same for all tokens issued by the same kubernetes api server. I also haven't added support for specifying the audience, but I would be happy to amend my PR if y'all would like those changes as well.

malnick:
@JWong101 Thank you for submitting this! 

I think it would be great if we could also include the custom aud configuration in this PR as well.

go.mod
```diff
@@ -3,17 +3,28 @@ module github.com/hashicorp/vault-plugin-auth-kubernetes
 go 1.12
 
 require (
+	github.com/Masterminds/semver v1.4.2-0.20190429190951-059deebd1619 // indirect
```
catsby: Would it be possible to remove the changes to `go.mod` and `go.sum` in this pull request? In general I believe we are moving to not including these in pull requests, and I don't see where any of these are either used here or required an update.

path_config.go
```diff
@@ -149,6 +159,8 @@ type kubeConfig struct {
 	CACert string `json:"ca_cert"`
 	// TokenReviewJWT is the bearer to use during the TokenReview API call
 	TokenReviewerJWT string `json:"token_reviewer_jwt"`
+
+	Issuer string `json:"issuer"`
```
catsby: We should add a code comment here similar to the other fields in the `kubeConfig` that details this is optional and a default assumed.

adammw:
@catsby @malnick has this made it to a release?

path_role.go
```diff
@@ -41,6 +41,10 @@ are allowed, both this and bound_service_account_namespaces can not be "*"`,
 					Description: `List of namespaces allowed to access this role. If set to "*" all namespaces
 are allowed, both this and bound_service_account_names can not be set to "*"`,
 				},
+				"audience": &framework.FieldSchema{
```
seh: Per my comment at https://github.com/terraform-providers/terraform-provider-vault/pull/601#discussion_r353856689, did you consider accepting a set of audiences here, requiring nonempty intersection with the audiences claimed by the token?

Issue #82, opened 2020-02-06
Default jwt token and ca cert when running in k8s ?


eonpatapon:
I'm running vault in kubernetes using the official helm chart, therefore ca.crt and k8s token are available in the vault pod in `/var/run/secrets/kubernetes.io/serviceaccount/`.

The helm chart also set up the vault serviceaccount with cluster role `system:auth-delegator` for token validation. So the pod token can be used a the `jwt_reviewer_token`.

Would it be acceptable for the plugin to check if `ca.crt` and `token` exists in `/var/run/secrets/kubernetes.io/serviceaccount/` and use them when enabling the plugin without specifying any ca cert or token.

This would greatly simplify my provisioning of vault when running in k8s.

I'm willing to work on a patch if it sounds good to you

reegnz:
I think this one was resolved with #83

tvoran:
@reegnz Yep, I think you're right! Support for this was included in 1.5, along with an option to control the behavior in 1.5.4: https://www.vaultproject.io/api-docs/auth/kubernetes#caveats

PR #83, opened 2020-02-24
Try reading the ca.crt and TokenReviewer JWT from the default service account

eh-steve:
This should help address https://github.com/hashicorp/vault-plugin-auth-kubernetes/issues/82

### DIFF
--- a/path_config.go
+++ b/path_config.go
@@ -7,6 +7,7 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"errors"
+	"io/ioutil"
 
 	"github.com/briankassouf/jose/jws"
 	"github.com/hashicorp/vault/sdk/framework"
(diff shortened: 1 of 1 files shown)


riuvshyn:
that would be amazing!!!

riuvshyn:
@tyrannosaurus-becks PTAL

riuvshyn:
@kalafut any chance this change can be considered to merge-in?
This PR simplifies k8s setup alot! really looking forward to this.

kalafut:
@riuvshyn Yes, we'll be adding it soon.

path_config.go
```diff
@@ -97,14 +98,26 @@ func (b *kubeAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Requ
 		return logical.ErrorResponse("no host provided"), nil
 	}
 
+	localCACert, _ := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/ca.crt")
+
+	localTokenReviewer, _ := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/token")
```
jasonodonnell, line 103: These errors should not be discarded since if they don't exist, it will cause problems.

eh-steve: If the errors are nil, the byte slice will also be nil, so will be converted into an empty string later. It’s only if both the default files and the configured cert data are empty that we have an error?

jasonodonnell: That's fair!

cyrus-mc:
@eh-steve 

I opened an issue that references this PR as the cause. 

I am unable to authenticate to external clusters (when vault is itself running in a cluster) due to the fact that in this change you default the tokenReviewerJWT to the POD token (if found) and one wasn't supplied in the auth configuration.

Issue #90, opened 2020-05-06
Make Issuer validation optional


riuvshyn:
According to JWT spec ISS validation could be optional but Audience must be validated.

Sometimes in k8s ISS can be updated so it will cause a massive outage of  services that using Vault.

PR #91, opened 2020-05-06
Adds optional validate_iss_claim config param to allow skip iss validation

riuvshyn:
Issuer validation should be optional and there should be a way to disable iss validation because currently if issuer will be changed all clients will not be able to log-in.

As per JWT spec issuer can be optional and it is optional in this plugin 
```
Description: "Optional JWT issuer. If no issuer is specified, then this plugin will use kubernetes.io/serviceaccount as the default issuer.",
```
However, with non-default issuer auth will fail due to validation as it will try to validate it against default `kubernetes.io/serviceaccount`.

Another way to avoid such issues is to stop trying to validate against default issuer `kubernetes.io/serviceaccount` and do validation only if issuer is explicitly set in the configuration 
but I am not sure about the backward compatibility of that approach.

fixes: https://github.com/hashicorp/vault-plugin-auth-kubernetes/issues/90

If that looks good I will open PRs to update Vault docs and terraform Vault provider to support this new param.

tests are passing:

```

$ make testacc

==> Checking that code complies with gofmt requirements...
go generate 
VAULT_ACC=1 go test -tags='vault-plugin-auth-kubernetes' $(go list ./... | grep -v /vendor/) -v  -timeout 45m
=== RUN   TestConfig_Read
--- PASS: TestConfig_Read (0.00s)
=== RUN   TestConfig
--- PASS: TestConfig (0.00s)
=== RUN   TestLogin
--- PASS: TestLogin (0.00s)
=== RUN   TestLogin_ECDSA_PEM
--- PASS: TestLogin_ECDSA_PEM (0.01s)
=== RUN   TestLogin_NoPEMs
--- PASS: TestLogin_NoPEMs (0.00s)
=== RUN   TestLoginSvcAcctAndNamespaceSplats
2020-06-04T12:12:36.007+0300 [ERROR] login unauthorized due to: JWT names did not match
--- PASS: TestLoginSvcAcctAndNamespaceSplats (0.00s)
=== RUN   TestAliasLookAhead
--- PASS: TestAliasLookAhead (0.00s)
=== RUN   TestLoginIssValidation
--- PASS: TestLoginIssValidation (0.00s)
=== RUN   TestLoginProjectedToken
=== RUN   TestLoginProjectedToken/normal
=== RUN   TestLoginProjectedToken/fail
=== RUN   TestLoginProjectedToken/projected-token
=== RUN   TestLoginProjectedToken/projected-token-expired
=== RUN   TestLoginProjectedToken/projected-token-invalid-role
--- PASS: TestLoginProjectedToken (0.00s)
    --- PASS: TestLoginProjectedToken/normal (0.00s)
    --- PASS: TestLoginProjectedToken/fail (0.00s)
    --- PASS: TestLoginProjectedToken/projected-token (0.00s)
    --- PASS: TestLoginProjectedToken/projected-token-expired (0.00s)
    --- PASS: TestLoginProjectedToken/projected-token-invalid-role (0.00s)
=== RUN   TestPath_Create
--- PASS: TestPath_Create (0.00s)
=== RUN   TestPath_Read
--- PASS: TestPath_Read (0.00s)
=== RUN   TestPath_Delete
--- PASS: TestPath_Delete (0.00s)
PASS
ok      github.com/hashicorp/vault-plugin-auth-kubernetes       0.843s
?       github.com/hashicorp/vault-plugin-auth-kubernetes/cmd/vault-plugin-auth-kubernetes      [no test files]
```

### DIFF
--- a/path_config.go
+++ b/path_config.go
@@ -57,12 +57,12 @@ extracted. Not every installation of Kuberentes exposes these keys.`,
 					Name: "JWT Issuer",
 				},
 			},
-			"validate_iss_claim": {
+			"disable_iss_validation": {
 				Type:        framework.TypeBool,
-				Description: "Optional JWT issuer validation. Allows to skip ISS validation, enabled by default",
-				Default:     true,
+				Description: "Optional JWT issuer validation. Allows to skip ISS validation.",
+				Default:     false,
 				DisplayAttrs: &framework.DisplayAttributes{
-					Name: "JWT Issuer Validation",
+					Name: "Disable JWT Issuer Validation",
 				},
 			},
 		},
@@ -87,11 +87,11 @@ func (b *kubeAuthBackend) pathConfigRead(ctx context.Context, req *logical.Reque
 		// Create a map of data to be returned
 		resp := &logical.Response{
 			Data: map[string]interface{}{
-				"kubernetes_host":    config.Host,
-				"kubernetes_ca_cert": config.CACert,
-				"pem_keys":           config.PEMKeys,
-				"issuer":             config.Issuer,
-				"validate_iss_claim": config.ValidateISSClaim,
+				"kubernetes_host":        config.Host,
+				"kubernetes_ca_cert":     config.CACert,
+				"pem_keys":               config.PEMKeys,
+				"issuer":                 config.Issuer,
+				"disable_iss_validation": config.DisableISSValidation,
 			},
 		}
 
@@ -109,7 +109,7 @@ func (b *kubeAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Requ
 	pemList := data.Get("pem_keys").([]string)
 	caCert := data.Get("kubernetes_ca_cert").(string)
 	issuer := data.Get("issuer").(string)
-	validateIssClaim := data.Get("validate_iss_claim").(bool)
+	disableIssValidation := data.Get("disable_iss_validation").(bool)
 	if len(pemList) == 0 && len(caCert) == 0 {
 		return logical.ErrorResponse("one of pem_keys or kubernetes_ca_cert must be set"), nil
 	}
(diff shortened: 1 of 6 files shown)


riuvshyn:
@tyrannosaurus-becks @kalafut WDYT?

kalafut:
I think the default should be flipped in order to preserve existing behavior. i.e. the parameter would allow you to opt-out of issuer validation.

riuvshyn:
@kalafut hmm... the default is set to `true` which means that validation is enabled by default.
Sorry if I get you confused with `skip` word in the PR title :) 
or you mean that I should rename `validate_iss_claim` -> `skip_iss_claim_validation` ?

Thanks for looking at this pr! :bow:

kalafut:
@riuvshyn For the config parameter handling you're correct, but existing configs (written earlier w/o this parameter), will be read as `ValidateISSClaim: false` which be different behavior when they run. Something like `disable_iss_validation` would match some existing `disable_*` parameters in Vault.

riuvshyn:
@kalafut updated PR.

riuvshyn:
@kalafut does it look good now?

path_config.go
```diff
@@ -57,6 +57,14 @@ extracted. Not every installation of Kuberentes exposes these keys.`,
 					Name: "JWT Issuer",
 				},
 			},
+			"disable_iss_validation": {
+				Type:        framework.TypeBool,
+				Description: "Optional JWT issuer validation. Allows to skip ISS validation.",
```
kalafut, line 62:
```suggestion
				Description: "Disable JWT issuer validation.",
```

path_config_test.go
```diff
@@ -261,10 +265,11 @@ func TestConfig(t *testing.T) {
...
-		PublicKeys: []interface{}{cert, cert2},
-		PEMKeys:    []string{testRSACert, testECCert},
-		Host:       "host",
-		CACert:     testCACert,
+		PublicKeys:           []interface{}{cert, cert2},
+		PEMKeys:              []string{testRSACert, testECCert},
+		Host:                 "host",
+		CACert:               testCACert,
+		DisableISSValidation: false,
 	}
```
kalafut, line 273: A test with `DisableISSValidation: true` would be good.

kalafut:
Looks good! One test request. Also can you run `go mod tidy`?

path_config_test.go
riuvshyn: added test.

path_config.go
riuvshyn: fixed.

riuvshyn:
@kalafut all done.

kalafut:
@riuvshyn This looks good for the config part. The one other test that would help would be to copy one of the successful TestLogin tests as a base and explicitly set some a `"issuer"` value in the config that will never match. The login call should fail. Then update the config to disable issuer validation and the login will then pass. That way we're explicitly testing the main focus of the PR.  Thanks!

riuvshyn:
@kalafut I've added `iss` validation login tests, PTAL.
also included test that covers default `iss` `kubernetes/serviceaccount` if no issuer defined explicitly.

Issue #93, opened 2020-07-26
bug: authentication to external k8s cluster fails


cyrus-mc:
Given the following situation:

* vault is running in cluster 1
* kubernetes authentication setup to external cluster (cluster 2) with no token_reviewer_jwt set

In this situation authentication will always fail due to the fact that Vault will default to using the token that vault itself is running under for the TokenReviewerJWT. As this has no authority in cluster2, authentication fails. 

Looking over the changes

https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/83

Is the PR that I believe broke this functionality due to the fact it added the following

```
if len(tokenReviewer) == 0 && len(localTokenReviewer) > 0 {
  tokenReviewer = string(localTokenReviewer)
}
```

As you can see if there is no tokenReviewer specified in the request it checks for the localTokenReviewer (found if running in a cluster) and defaults to that.

riuvshyn:
isn't this expected behaviour? this feature could work only with cluster where it runs otherwise you have to set `tokenReviewer` explicitly.

cyrus-mc:
@riuvshyn No that isn't true. If you don't specify the `tokenReviewer` it uses the JWT itself. 

As the documentation states

```
A service account JWT used to access the TokenReview API to validate other JWTs during login. If not set the JWT used for login will be used to access the API.
```

And as long as that JWT has the necessary permissions (system:auth-delegator).

I have verified this works as expected.

eh-steve:
I'm happy to open another PR to expose an extra config to specify whether to default to using the Vault deployment's SA JWT or the JWT to be verified as the auth delegator, but we'd need to decide which one we want as default behaviour.

I do find it a bit odd having to grant extra auth delegator privileges to service accounts which need to be authenticated, rather than the Vault SA being granted those privileges, but happy to take input?

tvoran:
Hi folks, the new `disable_local_ca_jwt` option that went out in vault 1.5.4 should allow you to disable use of the local token: https://www.vaultproject.io/api-docs/auth/kubernetes#disable_local_ca_jwt

PR #94, opened 2020-07-26
fix: use local token for TokenReviewer JWT only when using token for CA crt (#93)

cyrus-mc:
PR #83  added ability to read ca.crt and TokenReviewer JWT from the default service account under which Vault is running.

This ended up causing a breakage when trying to authenticate against external clusters in which no TokenReview JWT is specified in the configuration.

The changes in this PR make the assumption that the default service account TokenReviewer JWT should only be used if no CA certificate (or pem keys) are supplied and therefore they are read from the default service account.

### DIFF
--- a/path_config.go
+++ b/path_config.go
@@ -115,17 +115,16 @@ func (b *kubeAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Requ
 	caCert := data.Get("kubernetes_ca_cert").(string)
 	issuer := data.Get("issuer").(string)
 	disableIssValidation := data.Get("disable_iss_validation").(bool)
+	tokenReviewer := data.Get("token_reviewer_jwt").(string)
 	if len(pemList) == 0 && len(caCert) == 0 {
 		if len(localCACert) > 0 {
 			caCert = string(localCACert)
+			if len(tokenReviewer) == 0 && len(localTokenReviewer) > 0 {
+				tokenReviewer = string(localTokenReviewer)
+			}
 		} else {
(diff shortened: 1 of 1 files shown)


cyrus-mc:
A more elegant solution might be to determine somehow if the auth backend represents an in-cluster config. Key off the hostname (although that might not be 100% conclusive). Or add another option to the config that indicates this. 

As token_reviewer_jwt is an optional setting, assuming its value when it isn't set is not correct.

eh-steve:
I'd be happy with this, though we should probably update the docs to explain the default behaviour under each of these 3 cases:

- No CA and no token reviewer JWT set (default to local Vault SA credentials if present, otherwise error)
- CA set, no token reviewer JWT (defaults to using the JWT to be verified as the token reviewer JWT - requires giving your SAs which need to authenticate the `system:auth-delegator` role)
- Both CA and token reviewer JWT set, (explicitly uses the provided credentials)

cyrus-mc:
Any further discussion required on this? With the current state the docs don’t match the actual functionality.

yurifrl:
So, I'm getting this error 
```
failed to request new Vault token Error making API request.\n\nURL: PUT $VAULT_ADDR/v1/auth/$ENGINE/login\nCode: 400. Errors:\n\n* missing client token
```
Is that related?

tvoran:
Hi folks, thanks for the discussion and contribution here, but we decided to add an explicit option to control this behavior, which was released in vault 1.5.4; see https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/97 and https://github.com/hashicorp/vault/pull/9992 for the details.

Issue #107, opened 2021-06-02
Error parsing non-legacy JWT claims for service account tokens


ljupchokotev:
We are using the Vault CSI Driver which uses the Kubernetes Auth Method on Kubernetes v1.20. The CSI driver fails with the following message when trying to authenticate:

```
URL: POST https://vault/v1/auth/kubernetes/login
Code: 400. Errors:

* 1 error occurred:
	* error running lookahead function for mfa: could not parse UID from claims
```

This error comes from [here](https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/master/path_login.go#L156-L159). Vault expects a `kubernetes.io/serviceaccount/service-account.uid` claim in the JWT, which is only true for [legacy](https://github.com/kubernetes/kubernetes/blob/v1.20.7/pkg/serviceaccount/legacy.go) JWT claims in Kubernetes.

The Vault CSI driver uses the `api/v1/namespaces/<namespace>/serviceaccounts/<serviceaccount>/token` endpoint in Kubernetes to manually create a JWT for a service account and then uses that JWT to authenticate to Vault. The JWT created via this endpoint is using an [updated](https://github.com/kubernetes/kubernetes/blob/v1.20.7/pkg/serviceaccount/claims.go) version of the claims, so there is a new structure which Vault can't process.

Example decoded tokens:
1. Manually created JWT via the endpoint

```json
{
  "aud": [
    "aud"
  ],
  "exp": 1622641490,
  "iat": 1622637890,
  "iss": "iss",
  "kubernetes.io": {
    "namespace": "namespace",
    "serviceaccount": {
      "name": "vault",
      "uid": "4bbf3254-3461-476d-a1f8-0bc9cc00831b"
    }
  },
  "nbf": 1622637890,
  "sub": "system:serviceaccount:namespace:vault"
}
```

2. The automatically created token when creating the service account (from the `<service-account-name>-token` secret):
```json
{
  "iss": "kubernetes/serviceaccount",
  "kubernetes.io/serviceaccount/namespace": "namespace",
  "kubernetes.io/serviceaccount/secret.name": "vault-token-5qmzb",
  "kubernetes.io/serviceaccount/service-account.name": "vault",
  "kubernetes.io/serviceaccount/service-account.uid": "4bbf3254-3461-476d-a1f8-0bc9cc00831b",
  "sub": "system:serviceaccount:namespace:vault"
}
```

ljupchokotev:
[...]

PR #108, opened 2021-07-13
Fix Alias lookahead for new JWT private claims

tomhjp:
# Overview
The AliasLookahead callback is broken for non-legacy service account JWTs. The login operation itself has been updated to handle the new style of private JWT claims from Kubernetes, and this PR updates AliasLookahead to use the same struct for extracting the required UID out of either type of token.

The main consumer of this method is Vault Enterprise when any Sentinel EGPs are configured, as it populates the MFA entity information using the service account's UID.

# Design of Change
Login and AliasLookahead now use the same struct to extract UID and other information from JWTs.

#107

# Contributor Checklist
    - Changelog will be added to the Vault repo when the change is pulled in
- [x] Backwards compatible

### DIFF
--- a/path_login.go
+++ b/path_login.go
@@ -23,8 +23,6 @@ var (
 	// defaultJWTIssuer is used to verify the iss header on the JWT if the config doesn't specify an issuer.
 	defaultJWTIssuer = "kubernetes/serviceaccount"
 
-	uidJWTClaimKey = "kubernetes.io/serviceaccount/service-account.uid"
-
 	// errMismatchedSigningMethod is used if the certificate doesn't match the
 	// JWT's expected signing method.
 	errMismatchedSigningMethod = errors.New("invalid signing method")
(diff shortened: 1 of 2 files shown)


path_login.go
```diff
@@ -338,12 +343,12 @@ func (s *serviceAccount) namespace() string {
...
-	Namespace      string                      `mapstructure:"namespace"`
-	Pod            *projectedServiceAccountPod `mapstructure:"pod"`
-	ServiceAccount *projectedServiceAccountPod `mapstructure:"serviceaccount"`
+	Namespace      string     `mapstructure:"namespace"`
+	Pod            *reference `mapstructure:"pod"`
+	ServiceAccount *reference `mapstructure:"serviceaccount"`
 }
 
-type projectedServiceAccountPod struct {
+type reference struct {
```
benashz, line 351: nit: the name of this struct might be too generic,  I wonder if maybe `tokenMeta` could be any better?

benashz (approved):
Only one minor nit/suggestion, otherwise LGTM!

PR #110, opened 2021-09-02
Add support for setting an Alias' name from the k8s namespace and serviceaccount

benashz:
# Overview
This PR is a continuation of #103. Which added a new role configuration to store the `Alias.Name` in the form of `namespace/serviceaccount`. Prior to this change the `Alias.Name` would always be derived from the service account's token `uid`. 

# Design of Change
Add a new `role`s configuration `alias_name_source` to allow for setting one of the following naming formats:
- `<token.uid>` e.g. `53f8253e-6df4-4be7-adae-f566c524c02c` (remains the default)
- `<namespace>/<serviceaccount>`  e.g. `kube-system/coredns`

[ ] [PR #103](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pr/103)

# Contributor Checklist
[My Docs PR Link](link)
[ ] Add output for any tests not ran in CI to the PR description (e.g., acceptance tests)
[X] Backwards compatible

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"sort"
 	"strings"
 	"sync"
 
@@ -21,11 +20,15 @@ const (
 	aliasNameSourceDefault = aliasNameSourceSAToken
 )
 
-// map alias name source to its description
-var aliasNameSourceMap = map[string]string{
-	aliasNameSourceSAToken: "format: <token.uid>",
-	aliasNameSourceSAPath:  "format: <namespace>/<serviceaccount>",
-}
+var (
+	aliasNameSources = []string{aliasNameSourceSAToken, aliasNameSourceSAPath}
+	// map alias name source to its description
+	aliasNameSourceMap = map[string]string{
+		aliasNameSourceSAToken: "format: <token.uid>",
+		aliasNameSourceSAPath:  "format: <namespace>/<serviceaccount>",
+	}
+	errInvalidAliasNameSource = fmt.Errorf(`invalid alias_name_source, must be one of: %s`, strings.Join(aliasNameSources, ", "))
+)
 
 // kubeAuthBackend implements logical.Backend
 type kubeAuthBackend struct {
@@ -147,28 +150,23 @@ func (b *kubeAuthBackend) role(ctx context.Context, s logical.Storage, name stri
 }
 
 func validateAliasNameSource(source string) error {
-	sources := make([]string, len(aliasNameSourceMap))
-	i := 0
 	for s := range aliasNameSourceMap {
 		if s == source {
 			return nil
 		}
-		sources[i] = s
-		i++
 	}
-	sort.Strings(sources)
-	return fmt.Errorf(`invalid alias_name_source, must be one of: %s`, strings.Join(sources, ", "))
+	return errInvalidAliasNameSource
 }
 
 func getAliasNameSourceDesc() string {
-	var desc []string
-	for s, d := range aliasNameSourceMap {
+	var desc = make([]string, len(aliasNameSources))
+	for i, s := range aliasNameSources {
+		d := aliasNameSourceMap[s]
 		if s == aliasNameSourceDefault {
 			d = d + " [default]"
 		}
-		desc = append(desc, fmt.Sprintf("%q (%s)", s, d))
+		desc[i] = fmt.Sprintf("%q (%s)", s, d)
 	}
-	sort.Strings(desc)
 	return strings.Join(desc, ", ")
 }
 
--- a/backend.go
+++ b/backend.go
@@ -15,6 +15,8 @@ const (
 	configPath = "config"
 	rolePrefix = "role/"
 
+	// aliasNameSourceUnset provides backwards compatibility with preexisting roles.
+	aliasNameSourceUnset   = ""
 	aliasNameSourceSAToken = "sa_token"
 	aliasNameSourceSAPath  = "sa_path"
 	aliasNameSourceDefault = aliasNameSourceSAToken
@@ -142,15 +144,12 @@ func (b *kubeAuthBackend) role(ctx context.Context, s logical.Storage, name stri
 	if len(role.TokenBoundCIDRs) == 0 && len(role.BoundCIDRs) > 0 {
 		role.TokenBoundCIDRs = role.BoundCIDRs
 	}
-	if role.AliasNameSource == "" {
-		role.AliasNameSource = aliasNameSourceDefault
-	}
 
 	return role, nil
 }
 
 func validateAliasNameSource(source string) error {
-	for s := range aliasNameSourceMap {
+	for _, s := range aliasNameSources {
 		if s == source {
 			return nil
 		}
--- a/backend.go
+++ b/backend.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"sort"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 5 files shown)


path_login.go
```diff
@@ -95,11 +95,26 @@ func (b *kubeAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
 		return nil, errors.New("could not load backend configuration")
 	}
 
+	if err := validateAliasNameSource(role.AliasNameSource); err != nil {
```
calvn, line 98: Even though I see the good intent of having validation on `role.AliasNameSource` in here, this can actually cause errors on older entries that existed before this change since that value is not present for those roles.

backend.go
```diff
@@ -128,10 +139,39 @@ func (b *kubeAuthBackend) role(ctx context.Context, s logical.Storage, name stri
...
+	if role.AliasNameSource == "" {
+		role.AliasNameSource = aliasNameSourceDefault
+	}
 
 	return role, nil
 }
 
+func validateAliasNameSource(source string) error {
+	sources := make([]string, len(aliasNameSourceMap))
+	i := 0
```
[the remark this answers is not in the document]
benashz: fixed in https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/110/commits/4ecb7abf483096a5ab1e2d5d86fad18c17fcf5bf

path_login.go
```diff
@@ -95,11 +95,26 @@ func (b *kubeAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
...
+
 	serviceAccount, err := b.parseAndValidateJWT(jwtStr, role, config)
 	if err != nil {
 		return nil, err
 	}
 
+	var aliasName string
+	switch role.AliasNameSource {
+	case aliasNameSourceSAToken:
+		aliasName = serviceAccount.UID
```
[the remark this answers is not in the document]
benashz: Good catch! Fixed with https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/110/commits/23ce29fe0a96069475e173c17039b39ab8b7e18c

I will see if there is a better way to avoid this  confusion going forward.

path_login.go
benashz: I believe that the following update to `role()` mitigates this potential issue:

https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/23ce29fe0a96069475e173c17039b39ab8b7e18c/backend.go#L145-L147

benashz: addressed in https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/110/commits/40ecafed3570dcfd269833c7834264757cbed356

path_login.go
```diff
@@ -137,6 +142,17 @@ func (b *kubeAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
...
+	case aliasNameSourceSAPath:
+		return fmt.Sprintf("%s/%s", serviceAccount.Namespace, serviceAccount.Name), nil
+	default:
+		return "", fmt.Errorf("unknown alias_name_source %q", role.AliasNameSource)
+	}
+}
+
 // aliasLookahead returns the alias object with the SA UID from the JWT
 // Claims.
 func (b *kubeAuthBackend) aliasLookahead(_ context.Context, _ *logical.Request, data *framework.FieldData) (*logical.Response, error) {
```
calvn, line 158: We'll also have to update `aliasLookahead` to return the right alias format/value depending on the role's `alias_name_source` value as discussed offline.

path_role.go
```diff
@@ -302,6 +309,17 @@ func (b *kubeAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical
 		role.Audience = audience.(string)
 	}
 
+	if source, ok := data.GetOk("alias_name_source"); ok {
+		if err := validateAliasNameSource(source.(string)); err != nil {
+			return logical.ErrorResponse(err.Error()), nil
+		}
+		role.AliasNameSource = source.(string)
+	} else if role.AliasNameSource == aliasNameSourceUnset {
+		if s, ok := data.Schema["alias_name_source"]; ok {
```
calvn, line 318: I don't think you have to reach into `data.Schema` or fish out the default value through `s.Default` directly. `data.Get` will attempt to get the default value if it's unset or absent so we should be using that instead.

path_login.go
benashz: Yup, working on that now.

path_role.go
benashz: sweet, thanks!

benashz: fixed in https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/110/commits/40ecafed3570dcfd269833c7834264757cbed356

path_login.go
benashz: Support added in https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/110/commits/40ecafed3570dcfd269833c7834264757cbed356

backend.go
```diff
@@ -132,6 +148,27 @@ func (b *kubeAuthBackend) role(ctx context.Context, s logical.Storage, name stri
...
+func validateAliasNameSource(source string) error {
+	for _, s := range aliasNameSources {
+		if s == source {
+			return nil
+		}
+	}
+	return errInvalidAliasNameSource
+}
+
+func getAliasNameSourceDesc() string {
```
calvn, line 160: Since the map only contains 2 entries and the number of options is unlikely expand to be more than these, I wonder if it makes things easier to read if we were to have a multi-line string literal for description instead of building the string this way. It's not [too uncommon](https://github.com/hashicorp/vault/blob/e14d2038735c43fb1264b6db7158ca8c49125819/builtin/credential/aws/path_role.go#L92) for descriptions to be lengthy and declared as such.

calvn (approved):
A few nits, otherwise looks good!

path_login.go
```diff
@@ -137,12 +146,48 @@ func (b *kubeAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
...
+	if resp != nil {
+		return resp, nil
+	}
+
+	role, err := b.role(ctx, req.Storage, roleName)
+	if err != nil {
+		return nil, err
+	}
+	if role == nil {
+		return logical.ErrorResponse(fmt.Sprintf("invalid role name \"%s\"", roleName)), nil
```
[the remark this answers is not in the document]
benashz: I believe this was factored out from some previous code. I will update the other lines that to follow the suggested pattern.

path_login_test.go
```diff
@@ -561,35 +587,101 @@ func TestLoginSvcAcctAndNamespaceSplats(t *testing.T) {
...
-	}
-
-	req := &logical.Request{
-		Operation: logical.AliasLookaheadOperation,
-		Path:      "login",
-		Storage:   storage,
-		Data:      data,
-		Connection: &logical.Connection{
-			RemoteAddr: "127.0.0.1",
+	testCases := map[string]struct {
```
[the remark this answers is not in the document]
benashz: 👍

path_login.go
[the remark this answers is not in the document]
benashz: fixed in https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/110/commits/25d184220d8080fba882b1481db4b406771baf0a

backend.go
benashz: Seems reasonable, updating now.

benashz: Fixed in https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/110/commits/9eb89bd209fe72c4a4d05615eba02b1c9824f176

PR #112, opened 2021-09-17
fix: rename token alias source to sa_uid

benashz:
# Overview
The `token` alias source naming was a bit misleading since all alias naming attributes are derived from the contents of the service account token. This PR renames the parameter to `sa_uid` to more precisely define the source for the Alias' name.

# Design of Change
Trivial refactoring.

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -17,8 +17,8 @@ const (
 
 	// aliasNameSourceUnset provides backwards compatibility with preexisting roles.
 	aliasNameSourceUnset   = ""
-	aliasNameSourceSAUid   = "sa_uid"
-	aliasNameSourceSAPath  = "sa_path"
+	aliasNameSourceSAUid   = "serviceaccount_uid"
+	aliasNameSourceSAPath  = "serviceaccount_path"
 	aliasNameSourceDefault = aliasNameSourceSAUid
 )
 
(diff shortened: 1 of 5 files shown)


backend.go
```diff
@@ -17,14 +17,14 @@ const (
 
 	// aliasNameSourceUnset provides backwards compatibility with preexisting roles.
 	aliasNameSourceUnset   = ""
-	aliasNameSourceSAToken = "sa_token"
-	aliasNameSourceSAPath  = "sa_path"
-	aliasNameSourceDefault = aliasNameSourceSAToken
+	aliasNameSourceSAUid   = "serviceaccount_uid"
```
benashz, line 20: The renaming of the params was suggested by @tomhjp

backend.go
```diff
@@ -17,14 +17,14 @@ const (
...
+	aliasNameSourceSAPath  = "serviceaccount_path"
```
benashz, line 21: @tvoran, @tomhjp  wasn't that keen on this parameter's name what do  you think? The Alias' name is computed from the SA token's `namespace` and `serviceaccount` delimited by `/`

tvoran: Maybe just `serviceaccount_name`? Since UID vs name seems to be the big difference.

benashz: Thanks! Fixed in https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/112/commits/2fcb99aa83599cef93a5ef2ff6513004544e7c40

tvoran (approved):

PR #114, opened 2021-09-28
fix: validate JWT token on alias look ahead

benashz:
This PR adds support for validating that a request's JWT token contains the required claims for the bound role.  

# Overview
A high level description of the contribution, including:
Who the change affects or is for (stakeholders)?
What is the change? 
Why is the change needed?
How does this change affect the user experience (if at all)?

# Design of Change
Validate the JWT token against a known Vault role. If it is deemed to be invalid return an error.

Note: the token is not passed to Kubernetes token review API, as this is only needed for login.

### DIFF
--- a/path_login.go
+++ b/path_login.go
@@ -87,12 +87,9 @@ func (b *kubeAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
 		}
 	}
 
-	config, err := b.config(ctx, req.Storage)
+	config, err := b.loadConfig(ctx, req.Storage)
 	if err != nil {
 		return nil, err
-	}
-	if config == nil {
-		return nil, errors.New("could not load backend configuration")
 	}
 
 	serviceAccount, err := b.parseAndValidateJWT(jwtStr, role, config)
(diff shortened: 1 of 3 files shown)


scripts/update_deps.sh
```diff
@@ -1,37 +0,0 @@
-#!/bin/sh
```
calvn, line 1: I wonder if the script changes should be a separate PR to make the backport/cherry-picking process easier.

benashz: Good point :)

benashz: Fixed.

tvoran (approved):

path_login.go
```diff
@@ -190,15 +187,12 @@ func (b *kubeAuthBackend) aliasLookahead(ctx context.Context, req *logical.Reque
...
-	parsedJWT, err := jws.ParseJWT([]byte(jwtStr))
+	config, err := b.loadConfig(ctx, req.Storage)
 	if err != nil {
 		return nil, err
 	}
 
-	// Decode claims into a service account object
-	sa := &serviceAccount{}
-	err = mapstructure.Decode(parsedJWT.Claims(), sa)
+	sa, err := b.parseAndValidateJWT(jwtStr, role, config)
```
tomhjp, line 195: Is it worth leaving a comment behind to ensure the intent is clear and help prevent similar bugs in future patches? It seems like an easy mistake to make.

tomhjp (approved):

path_login.go
benashz: Fixed in. https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/114/commits/aec50977ce5b5f1aab03c38c4153abec583105fe

tomhjp (approved):

PR #115, opened 2021-09-28
maint: misc. updates for build and CI

benashz:
- use golang:1.16.7 in CI
- remove vendor references from build scripts
- remove obsolete update_deps.sh build script


Makefile
```diff
@@ -1,5 +1,5 @@
 TOOL?=vault-plugin-auth-kubernetes
-TEST?=$$(go list ./... | grep -v /vendor/)
+TEST?=$$(go list ./...)
```
tvoran, line 2: Looks like we could remove the `| grep -v vendor` from GOFMT_FILES too? 

https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/e108b81f10613b52866cc9f7659b94d3154c481a/Makefile#L8

.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: docker.mirror.hashicorp.services/circleci/golang:1.15
+      - image: docker.mirror.hashicorp.services/circleci/golang:1.16.7
```
tvoran, line 5: FWIW I think 1.16.8 is out now. Do we need to pin to a specific patch version or would `golang:1.16` suffice?

tvoran (approved):

.circleci/config.yml
benashz: I was following Vault core, they pin to the patch. I am okay to pin to the minor, but then we will have variation if ever a build is rerun.

Makefile
benashz: Good  catch, missed that one.

benashz: Fixed in https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/115/commits/441d1b4301a351fc8f3b0c73eb5746ff4553a791

tvoran (approved):

tomhjp (approved):

PR #140, opened 2022-02-18
Use golang-jwt/jwt/v4 to be inline with vault.

benashz:
# Overview
This PR updates the tests to use `github.com/golang-jwt/jwt/v4` instead of the `incompatible` v3 module. This is inline with some recent changes to vault and its indirect dependencies introduced in https://github.com/hashicorp/vault/pull/14138

### DIFF
--- a/path_login_test.go
+++ b/path_login_test.go
@@ -11,10 +11,10 @@ import (
 	"time"
 
 	"github.com/briankassouf/jose/jws"
-	// TODO: using github.com/golang-jwt/jwt for tests only,
+	// TODO: using github.com/golang-jwt/jwt/v4 for tests only,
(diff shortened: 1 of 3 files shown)


path_login_test.go
```diff
@@ -11,10 +11,10 @@ import (
 	"time"
 
 	"github.com/briankassouf/jose/jws"
-	// TODO: using github.com/golang-jwt/jwt for tests only,
+	// TODO: using github.com/golang-jwt/jwt/v4 for tests only,
 	// as a part of moving away from jose we should consider standardizing
```
tvoran, line 15:
```suggestion
	// as a part of moving away from the jose fork we should consider standardizing
```

tvoran (approved):

PR #142, opened 2022-03-18
Cache httpClient for connection re-use

imthaghost:
# Overview
 Currently, the `Review` [function](https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/21abc8d40de83a6a62584e606aa5dc5a5ae0ed5a/token_review.go#L50) for this backend creates a new HTTP client every time it gets called. This can be quite expensive and does not allow the backend to re-use connections. 

# Design of Change
In this implementation, the `kubeAuthBackend` struct stores and instantiates a `DefaultPooledHTTPClient` for connection re-use. The client is then passed by reference down to the `reviewFactory` for use in the Kubernetes API. This will ensure that we only create and use one HTTP client.


backend.go
```diff
@@ -103,6 +108,11 @@ func Backend() *kubeAuthBackend {
 		),
 	}
 
+	// create and store a new DefaultPooledClient for connection reuse
+	if b.httpClient == nil {
+		b.httpClient = cleanhttp.DefaultPooledClient()
+	}
```
tvoran, line 114: `b.httpClient` will always be nil here, right? I wonder if this check would fit better in the Review function.

tomhjp: Alternatively, maybe just set it directly in the `&framework.Backend{...}` struct initialisation above?

token_review.go
```diff
@@ -174,13 +174,14 @@ func parseResponse(resp *http.Response) (*authv1.TokenReview, error) {
 
 // mock review is used while testing
 type mockTokenReview struct {
-	saName      string
-	saNamespace string
-	saUID       string
+	saName            string
+	saNamespace       string
+	saUID             string
+	checkPooledClient bool
```
tomhjp, line 180: What's the status of this PR? Looks like this is currently unused, and I'm not sure that you'll need it, unless it causes you difficulties while writing the test.

backend.go
imthaghost: @tvoran You are right! In commit [2d36b73](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/142/commits/2d36b73054af33d72d407b211b39f7f6bf2768c4) I removed the nil check from the backend struct and added the nil check inside of service account where token review has access to the backend.

imthaghost: @tomhjp I don't believe I need to set client inside of `&framework.Backend{...}` because the service lookup function already has access to the `kubAuthBackend` struct so all I had to do was pass the client as a reference to the only caller which is the Review function.

token_review.go
imthaghost: Yes ideally I wanted to check or have some indication we are using the `PooledClient` I think it does cause a lot of complications as the only way of getting access to that information is by looking at the `Transport`. Do you have any other suggestions on how to make sure the connection is being reused? The only option I can think of is using client trace to check whether the request's underlying tcp connection was re-used. Example: https://golang.cafe/blog/golang-httptrace-example.html

backend.go
tomhjp: Oh ignore me, I misread the code 👍 what you've got makes sense.

token_review.go
```diff
@@ -189,7 +187,7 @@ func mockTokenReviewFactory(name, namespace, UID string) tokenReviewFactory {
 	}
 }
 
-func (t *mockTokenReview) Review(ctx context.Context, cjwt string, aud []string) (*tokenReviewResult, error) {
+func (t *mockTokenReview) Review(ctx context.Context, cjwt string, aud []string, client *http.Client) (*tokenReviewResult, error) {
```
tomhjp, line 190: I think the suggestion elsewhere was essentially that you could make this mock function return an error if the client is not configured correctly. i.e. something like:

```go
httpTransport, ok := client.Transport.(*http.Transport)
if !ok {
	return nil, errors.New("failed to check whether DisableKeepAlives is false as Transport is not *http.Transport")
}
if httpTransport.DisableKeepAlives {
	return nil, errors.New("expected DisableKeepAlives to be false but was true")
}
```

An alternative might be to add a similar check more directly on the returned value from `Backend()` in a new test.

token_review.go
tomhjp: I've left a comment on the mock function code in response to this.

backend.go
```diff
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/hashicorp/go-cleanhttp"
```
kalafut, line 8: Please run `goimports`

token_review.go
imthaghost: Updated in https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/142/commits/246c3f85023b0bcfd46ebc3f667658b274e74cf5

backend.go
imthaghost: Updated in https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/142/commits/246c3f85023b0bcfd46ebc3f667658b274e74cf5

path_login.go
```diff
@@ -102,8 +104,14 @@ func (b *kubeAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
 		return nil, err
 	}
 
+	// ensure the http client is not empty
+	if b.httpClient == nil {
+		b.httpClient = cleanhttp.DefaultPooledClient()
+	}
+
```
tomhjp, line 111: I don't think this is necessary. We should never instantiate the plugin without using the `Backend` function

imthaghost: That makes sense! Changed in [dac4c83](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/142/commits/dac4c8339f2b90997f7148f3832e0df1ad5ae525)

token_review.go
```diff
@@ -45,9 +44,7 @@ func tokenReviewAPIFactory(config *kubeConfig) tokenReviewer {
 	}
 }
 
-func (t *tokenReviewAPI) Review(ctx context.Context, jwt string, aud []string) (*tokenReviewResult, error) {
-
-	client := cleanhttp.DefaultClient()
+func (t *tokenReviewAPI) Review(ctx context.Context, jwt string, aud []string, client *http.Client) (*tokenReviewResult, error) {
 
 	// If we have a CA cert build the TLSConfig
 	if len(t.config.CACert) > 0 {
```
calvn, line 50: I realized that we might get into concurrency issues here now that we're sharing the same `client` across requests (since this gets invoked on logins). Concurrent modification of the client in here might be problematic, so we'd have guard updates with a mutex write lock (and subsequently grab a read lock during the `Review` call). Now that we're storing the client in memory, we shouldn't have to set this on every review call; we can set the transport on instantiation of the client and on config updates/deletes.

backend.go
```diff
@@ -103,6 +108,9 @@ func Backend() *kubeAuthBackend {
 		),
 	}
 
+	// Set a default http client.
+	b.httpClient = cleanhttp.DefaultPooledClient()
+
```
calvn, line 113: Now that we've moved setting the client's `TLSClientConfig` on config write, we should also consider setting it when we're instantiating the client so that it is also set whenever backend restarts.

token_review.go
```diff
@@ -29,7 +28,7 @@ type tokenReviewResult struct {
 
 // This exists so we can use a mock TokenReview when running tests
 type tokenReviewer interface {
-	Review(context.Context, string, []string) (*tokenReviewResult, error)
+	Review(context.Context, string, []string, *http.Client) (*tokenReviewResult, error)
```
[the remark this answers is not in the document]
imthaghost: Adjusted in https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/142/commits/91bca5a546e8818e49cf51cc09beb752cb8c8cc9

backend.go
```diff
@@ -103,6 +110,15 @@ func Backend() *kubeAuthBackend {
...
 	}
 
+	// Set a default HTTP client
+	b.httpClient = cleanhttp.DefaultPooledClient()
+
+	// Set the HTTP client's TLS config
+	b.httpClient.Transport.(*http.Transport).TLSClientConfig = &tls.Config{
+		RootCAs:    x509.NewCertPool(),
+		MinVersion: tls.VersionTLS12,
+	}
```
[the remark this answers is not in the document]
imthaghost: When loading the config we check if there is a local CA cert unless it was stored in the config. I've added an additional line in [5c1e183](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/142/commits/5c1e183207a8b8b3fc8946fb6565b6bfb42ccd76) that will build the TLS config and build the client.

backend.go
```diff
@@ -164,11 +174,22 @@ func (b *kubeAuthBackend) loadConfig(ctx context.Context, s logical.Storage) (*k
...
+	} else {
+		certPool := x509.NewCertPool()
+		certPool.AppendCertsFromPEM([]byte(config.CACert))
+
+		tlsConfig := &tls.Config{
+			MinVersion: tls.VersionTLS12,
+			RootCAs:    certPool,
+		}
+
+		b.httpClient.Transport.(*http.Transport).TLSClientConfig = tlsConfig
```
calvn, line 192: There might be two issues if we set this on `loadConfig`:

1. We'd have to grab a write lock during this portion of the call since we're modifying the backend's HTTP client. However, `loadConfig` acts more like a "read" operation that retrieves the config from storage.
2. We'd get to a state where we're now setting/updating the client on every `loadConfig` call, which happens on every login (and aliasLookahead) call.

We could get around issue #2 by using a sentinel value to track whether we've set this already, and only set it once during the lifecycle of the backend object. However, it might be simpler to set it when the backend is instantiated (i.e the `Backend()`) or within a backend's `InitializeFunc`.

imthaghost: @tomhjp I also wanted to loop you in on [d1e904b](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/142/commits/d1e904bd7f5f955105a51bf0d9464c5c52ca888a) I added an `initialize` method to the `kubeAuthBackend` struct. The reasoning for this is because we don't have access to the config without creating additional logic inside the `Backend()` func which is primarily just used for setup/configuration it made more sense to try and load the config on initialization when the auth backend is mounted. This still allows us to continue as normal if a config doesn't exist and if a config and a trusted CA cert exist we can load it into the HTTP client. @calvn showed me a great example of this flow being used in the AWS auth method [here](https://github.com/hashicorp/vault/blob/86c1cdf9e61897dee14203c51d6ad3d971cfda1a/builtin/credential/aws/backend.go#L189).

backend.go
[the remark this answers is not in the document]
imthaghost: Addressed in https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/142/commits/d1e904bd7f5f955105a51bf0d9464c5c52ca888a

path_login.go
```diff
@@ -103,8 +103,11 @@ func (b *kubeAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
 		return nil, err
 	}
 
+	b.l.RLock()
```
tomhjp, line 106: Don't we already have the read lock from line 69 above?

backend.go
```diff
@@ -101,14 +108,47 @@ func Backend() *kubeAuthBackend {
...
 }
 
+// initialize is used to handle the state of config values just after a plugin has been mounted
+func (b *kubeAuthBackend) initialize(ctx context.Context, req *logical.InitializationRequest) error {
+	// Try to load the config on init
+	config, err := b.loadConfig(ctx, req.Storage)
+	if err != nil {
+		return err
+	}
+	if config == nil {
```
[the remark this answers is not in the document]
imthaghost: Updated in https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/142/commits/c1230008498ef59342ab8495ffe33176de931ace. This allows other functions to return an error if the config is nil and this will allow us in our initialize fucntion to check if we have a nil config value and just continue as normal.

path_login.go
imthaghost: You are right! However, I don't believe we would run into a deadlock I could be mistaken.

path_config.go
```diff
@@ -154,6 +156,22 @@ func (b *kubeAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Requ
 		DisableLocalCAJwt:    disableLocalJWT,
 	}
 
+	b.l.Lock()
+	defer b.l.Unlock()
+
+	// If we have a CA cert build the TLSConfig
+	if len(config.CACert) > 0 {
```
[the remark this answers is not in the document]
imthaghost: If I am understanding this properly: we have to handle the case where `disable_local_ca_jwt == false` and the case where `disable_local_ca_jwt == true`. Currently, we are loading the CA cert from the `kubernetes_ca_cert` path and building our TLS config using that CA cert.  But not handling the local CA cert.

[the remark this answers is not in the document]
tomhjp: That's correct 👍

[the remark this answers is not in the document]
imthaghost: In https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/142/commits/8de6823103f788c6bde6a8cd4783f6e7abcec9d1 we check if `disable_local_ca_jwt == true' and continue as we did previously in addition we have an else statement that will use the `localCACertPath` to build the TLS config.

path_config.go
```diff
@@ -154,6 +155,27 @@ func (b *kubeAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Requ
...
+	tlsConfig := &tls.Config{
+		MinVersion: tls.VersionTLS12,
+	}
+	if disableLocalJWT {
+		certPool.AppendCertsFromPEM([]byte(config.CACert))
+		tlsConfig.RootCAs = certPool
+
+		b.httpClient.Transport.(*http.Transport).TLSClientConfig = tlsConfig
+	} else {
+		certPool.AppendCertsFromPEM([]byte(localCACertPath))
```
tomhjp, line 173: Have you tested this code path yet?

imthaghost: Yup tested by forcing `disable_local_ca_jwt == false`  and logging what exists inside the TLS config when the auth backend is mounted. I could be mistaken but it seems like I also don't need to specify `localCACertPath` because the local CA cert is loaded into the config unless we specify we want to overwrite the CA Cert with the `kubernetes_ca_cert` path.

backend.go
calvn: It looks like one of the commits reverted your update, can we get this fixed once more?

backend.go
```diff
@@ -144,10 +183,10 @@ func (b *kubeAuthBackend) loadConfig(ctx context.Context, s logical.Storage) (*k
 	if err != nil {
 		return nil, err
 	}
+	// We know the config is empty so exit early
 	if config == nil {
-		return nil, errors.New("could not load backend configuration")
+		return config, nil
```
calvn, line 188: Another place that we're calling `loadConfig` is: 
https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/9a0d8f05f80dfd6a4053b7a303f1afe23ac11138/path_login.go#L88-L93

We should also nil-check in there to avoid panics. Would it be good to also call out this behavior in `loadConfig`'s doc-string so future users know that they should nil-check the config? Are there any test breakages that would result from this change (I see `loadConfig` being called directly a handful of times in tests)?

path_config.go
calvn: What did you observe in your log output? Was the auth method able to establish a connection to k8s with the flag set to `false`?

> local CA cert is loaded into the config

Are we sure that the  (local k8s) CA cert is loaded into the config at this point in time, i.e. a cert that was not provided through `caCert`?

path_login.go
```diff
@@ -411,8 +417,8 @@ type k8sObjectRef struct {
 
 // lookup calls the TokenReview API in kubernetes to verify the token and secret
 // still exist.
-func (s *serviceAccount) lookup(ctx context.Context, jwtStr string, tr tokenReviewer) error {
-	r, err := tr.Review(ctx, jwtStr, s.Audience)
+func (s *serviceAccount) lookup(ctx context.Context, jwtStr string, tr tokenReviewer, client *http.Client) error {
```
calvn, line 420: It'd be nice to also have the `*http.Client` go right after `context.Context` in here, similar to how we have it for the `tokenReviewAPI.Review` method.

backend.go
imthaghost: Updated in [139b7d6](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/142/commits/139b7d6100e9856ac19f1f010f6fbd85fb3fe249)

path_login.go
imthaghost: Updated in https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/142/commits/e1765a3788f49f37fbf426d7ad8025613496d7a5

backend.go
imthaghost: Updated in [e1765a3](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/142/commits/e1765a3788f49f37fbf426d7ad8025613496d7a5)

path_config.go
imthaghost: > What did you observe in your log output? Was the auth method able to establish a connection to k8s with the flag set to `false`?
> 
> > local CA cert is loaded into the config
> 
> Are we sure that the (local k8s) CA cert is loaded into the config at this point in time, i.e. a cert that was not provided through `caCert`?

The CA Cert wasn't being loaded in at the time. This is why in https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/142/commits/f5470c1fbab63e24c02ab0119c959b5090c5ca49 I changed the statement to read the local CA cert and passes it into the function `AppendCertsFromPEM`. In this change, we also handle the case when the config CA cert exists so we can load the cert directly from the config.

calvn (approved):
Looks good, thanks for addressing all of the incoming feedback!

PR #147, opened 2022-04-04
1.10 Backport: Fix error code when using the wrong service account (#144)

tomhjp:
Backports #144 to 1.10.

* Fix error code when using the wrong service account
* Add integration tests
* Use helm install + hostPort patch instead of raw manifest
* Add unit tests
* Remove newly redundant CircleCI config


.github/workflows/tests.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - uses: actions/checkout@v2
+      - run: make fmtcheck
+
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.18.0
```
tvoran, line 18: Not sure that it matters too much, but I think Vault 1.10 is still on go 1.17.x

tvoran (approved):

.github/workflows/tests.yaml
tomhjp: Good point, may as well match it up for the release branch

Issue #30, opened 2018-06-15
Error validating serviceaccount token when "pem_keys" are specified in the configuration


primeroz:
Hi,

I have been trying to use the kubernetes authentication plugin and i was successful until i tried to use the "pem_keys" to validate the signature of the serviceaccount provided by the client pod as documented [here](https://www.vaultproject.io/api/auth/kubernetes/index.html#pem_keys)

I am running vault 0.10.1 and kubernetes 1.9.8 , i am using ECDSA Key to sign the serviceaccounts and i extracted the public key from the key using `openssl ec -in /etc/kubernetes/ssl/serviceaccount.key -pubout` 

I then updated the kubernetes auth config doing
```
 vault write auth/kubernetes/config kubernetes_host=URL token_reviewer_jwt=VAULT_SERVICE_TOKEN kubernetes_ca_cert=@/etc/ssl/certs/kubernetes-ca.pem pem_keys="PUBLIC_KEY_FROM_OPENSSL_COMMAND"
```

Doing a read show the key succesfully loaded in the vault config for kubernetes auth

Whenever i try to login from my client though i get an error 
```
~ $ curl -X POST -d @/dev/shm/payload.json -k https://vault:8200/v1/auth/kubernetes/login                                                           
{"errors":["asn1: structure error: length too large"]}
```

The very same command work as long as i don't have any "pem_keys" defined 

I also verified that the signature is correct for the public key using https://jwt.io/ 

Is there anything i am maybe missing in my configuration or anything else i am doing wrong ?

jefferai:
What is the value of `PUBLIC_KEY_FROM_OPENSSL_COMMAND`?

primeroz:
Hi Jeff.

If i understood correctly you want to see the output of the ssl command pubout key 
```
sudo openssl ec -in /etc/kubernetes/ssl/serviceaccount.key -pubout
read EC key
writing EC key
-----BEGIN PUBLIC KEY-----
MHYwEAYHKoZIzj0CAQYFK4EEACIDYgAEzpOY+XvrouGrV84L5BiIVg8/x85xsav6
cbDHnTXcwiTP830Aeg7w18dS4RylgiWFx8+E/BnKD1VSdHCfWFsskAQXAcRMIPbN
scOKeyb+rupBpled89vbPNlKd0CUg+Vo
-----END PUBLIC KEY-----
```

primeroz:
[...]

PR #149, opened 2022-04-12
VAULT-2003 WIP switch to cap + jose

swenson:
# Overview

* Remove `golang-jwt` library direct dependency
* Use `cap` for checking JWT claims
* Add additional integration tests to cover more cases
* Switch to gofumpt to match vault
* Clean up token handing and signing in tests

This mostly pays down technical debt and should hopefully make it easier to work with our JWT code moving forward.

This should cause minimal user-facing changes (notably there might be slight differences in the status code messages when signatures are invalid).

* #30
* #33
* Replaces #148

# Contributor Checklist
* [x] Add relevant docs to upstream Vault repository, or sufficient reasoning why docs won’t be added yet (no user-facing changes made)
* [x] Add output for any tests not ran in CI to the PR description (eg, acceptance tests)
* [x] Backwards compatible


integrationtest/integration_test.go
```diff
@@ -18,26 +20,56 @@ import (
 //   - A Kubernetes cluster in which:
 //       - it can use the `test` namespace
 //       - Vault is deployed and accessible
+//       - There is a serviceaccount called test-token-reviewer-account with access to the TokenReview API
```
tvoran, line 23: Interesting, I think the default token in the vault-0 pod should already have access to the TokenReivew API: https://github.com/hashicorp/vault-helm/blob/4ae52c8bd3d3f4d303196ecb455ae07d9a943ba8/templates/server-clusterrolebinding.yaml#L17-L25

swenson:
I'll close this for now in favor of #148 but keep the branch for if we want to circle back to it later.

integrationtest/integration_test.go
```diff
@@ -51,29 +83,97 @@ func TestUnauthorizedServiceAccountErrorCode(t *testing.T) {
...
+	defer deferred()
+
+	_, err := client.Logical().Write("auth/kubernetes/login", map[string]interface{}{
+		"role": "test-role",
+		"jwt":  os.Getenv("KUBERNETES_JWT"),
+	})
+	if err != nil {
+		t.Fatalf("Expected successful login but got: %v", err)
+	}
+}
```
tvoran, line 150: It would be good to also test a login where `token_reviewer_jwt` is blank on the config, so the incoming `jwt` is also used to auth with the TokenReview API.

tvoran (commented):
Oops, didn't realize you closed this 😃

PR #150, opened 2022-04-21
Remove cert and pem enforcement

imthaghost:
# Overview
Some users didn't find it necessary for Vault to enforce the CA cert or PEM keys. Examples can be found in the related issues.

# Design of Change
We add false booleans to the `kubernetes_ca_cert` and` pem_keys` fields. While the `required`  field on the `FieldSchema` struct is deprecated we added this just for future reference and documentation. We also remove checks that enforced `kubernetes_ca_cert`  to be present instead we just default to the local CA cert if `kubernetes_ca_cert` is not set which will return an error of `x509: certificate signed by unknown authority` if the user did supply an appropriate CA cert. 

# Related Issues
[Issue #1234](https://github.com/hashicorp/vault/issues/1234)
[Issue #88](https://github.com/hashicorp/vault-plugin-auth-kubernetes/issues/88)

# Docs
I don't believe docs needed to be added since this is implied in the example below.

Screen Shot 2022-04-21 at 4 11 55 PM

### DIFF
--- a/path_config.go
+++ b/path_config.go
@@ -178,7 +178,7 @@ func (b *kubeAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Requ
 
 		b.httpClient.Transport.(*http.Transport).TLSClientConfig = tlsConfig
 	}
-	
+
 	var err error
 	for i, pem := range pemList {
 		config.PublicKeys[i], err = parsePublicKeyPEM([]byte(pem))
--- a/path_config.go
+++ b/path_config.go
@@ -57,7 +57,9 @@ extracted. Not every installation of Kubernetes exposes these keys.`,
 				DisplayAttrs: &framework.DisplayAttributes{
 					Name: "Service account verification keys",
 				},
+				Required: false,
 			},
+
 			"issuer": {
 				Type:       framework.TypeString,
 				Deprecated: true,
--- a/path_config.go
+++ b/path_config.go
@@ -37,6 +37,7 @@ func pathConfig(b *kubeAuthBackend) *framework.Path {
 				DisplayAttrs: &framework.DisplayAttributes{
 					Name: "Kubernetes CA Certificate",
 				},
+				Required: false,
 			},
 			"token_reviewer_jwt": {
 				Type: framework.TypeString,
@@ -141,10 +142,6 @@ func (b *kubeAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Requ
 		}
 	}
 
-	if disableLocalJWT && caCert == "" {
-		return logical.ErrorResponse("kubernetes_ca_cert must be given when disable_local_ca_jwt is true"), nil
-	}
-
 	config := &kubeConfig{
 		PublicKeys:           make([]interface{}, len(pemList)),
 		PEMKeys:              pemList,
@@ -181,7 +178,7 @@ func (b *kubeAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Requ
 
 		b.httpClient.Transport.(*http.Transport).TLSClientConfig = tlsConfig
 	}
-
+	
 	var err error
 	for i, pem := range pemList {
 		config.PublicKeys[i], err = parsePublicKeyPEM([]byte(pem))


tomhjp (commented):
Please can we get some unit test coverage? It's harder to test in an automated way, but I'd also like to ensure we at least manually test using the OS trust store. LMK if you'd like any help with that and I can go into more detail.

path_config.go
```diff
@@ -56,7 +57,9 @@ extracted. Not every installation of Kubernetes exposes these keys.`,
 				DisplayAttrs: &framework.DisplayAttributes{
 					Name: "Service account verification keys",
 				},
+				Required: false,
 			},
+
```
tvoran, line 62:
```suggestion
```

imthaghost: Removed white space [0027a14](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/150/commits/0027a145e05f0c43ae4f26f6bc1820b43cd71bcc)

path_config_test.go
```diff
@@ -535,6 +535,62 @@ func TestConfig_LocalJWTRenewal(t *testing.T) {
...
+			setupInClusterFiles: true,
+			expected: &kubeConfig{
+				PublicKeys:           []interface{}{},
+				PEMKeys:              []string{},
+				Host:                 "host",
+				CACert:               testLocalCACert,
+				TokenReviewerJWT:     testLocalJWT,
+				DisableISSValidation: true,
+				DisableLocalCAJwt:    true,
+			},
```
tvoran, line 557: It looks like this test is failing; is the expected kubeConfig correct here?

imthaghost: Updated in [a0dfafe](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/150/commits/a0dfafee6502d974689c2ce60a30265583d37d1f) to test the right code path. The `localCACert` is added to the OS trust store however we can remove this entirely and generate our own `self-signed cert` and add it to the OS trust sore when a cert is not supplied.

path_config_test.go
```diff
@@ -535,6 +535,65 @@ func TestConfig_LocalJWTRenewal(t *testing.T) {
...
+			expected: &kubeConfig{
+				PublicKeys:           []interface{}{},
+				PEMKeys:              []string{},
+				Host:                 "host",
+				CACert:               testLocalCACert,
+				TokenReviewerJWT:     testLocalJWT,
+				DisableISSValidation: true,
+				DisableLocalCAJwt:    false,
+			},
+		},
```
tvoran, line 561: Is this test different than "no CA or JWT, default to local" in `TestConfig_LocalCaJWT()`? IIRC `disable_local_ca_jwt` defaults to false, and `kubernetes_ca_cert` defaults to `""`.
https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/a0dfafee6502d974689c2ce60a30265583d37d1f/path_config_test.go#L361-L375

tvoran, line 561: Is this test different than "no CA or JWT, default to local" in `TestConfig_LocalCaJWT()`? IIRC `disable_local_ca_jwt` defaults to false, and `kubernetes_ca_cert` defaults to `""`.
https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/a0dfafee6502d974689c2ce60a30265583d37d1f/path_config_test.go#L361-L375

path_config.go
```diff
@@ -33,10 +33,11 @@ func pathConfig(b *kubeAuthBackend) *framework.Path {
 
 			"kubernetes_ca_cert": {
 				Type:        framework.TypeString,
-				Description: "PEM encoded CA cert for use by the TLS client used to talk with the API.",
+				Description: "Optional PEM encoded CA cert for use by the TLS client used to talk with the API.",
 				DisplayAttrs: &framework.DisplayAttributes{
 					Name: "Kubernetes CA Certificate",
 				},
+				Required: false,
```
calvn, line 40: This is not immediately apparent, but this field does not have any [actual effect](https://github.com/hashicorp/vault/blob/5c4b49c457774b58fbb065dc7040237648f0e75c/sdk/framework/backend.go#L673) on the enforcement of whether a field schema is required or not. The `Required` bool here is purely for letting consumers of openapi (e.g. the UI) know whether the field should be treated as a required. 

Since the zero-value of this field is `false`, specifying this might not have any effect at all, even for openapi evaluation.

The removal of the check further below, is the right thing to do to remove the cert requirement though.

path_config.go
```diff
@@ -161,11 +159,11 @@ func (b *kubeAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Requ
 
 	// Determine if we load the local CA cert or the CA cert provided
 	// by the kubernetes_ca_cert path into the backend's HTTP client
-	certPool := x509.NewCertPool()
 	tlsConfig := &tls.Config{
 		MinVersion: tls.VersionTLS12,
 	}
 	if disableLocalJWT || len(caCert) > 0 {
```
tomhjp, line 165: How come we don't use system certs if `disableLocalJWT` is true?

tomhjp, line 165: This is a more idiomatic check for empty strings:

```suggestion
	if disableLocalJWT || caCert != "" {
```

path_config.go
imthaghost: The `Required` bool was added in just for future reference and documentation (lettings consumers know how the field is treated as mentioned)

rinormaloku:
@imthaghost and @tomhjp can we merge this?

PR #151, opened 2022-05-02
Switch from golang-jwt to cap+jose

swenson:
# Overview

This caused stricter token checking around for Issued At and Expiration,
so we cleaned up the tokens in the tests to be stored in plain JSON and
signed as part of the test initialization to make it easier to see what
is going on. This also means that the tests won't start mysteriously
failing in 2030. :)

We also updated to the latest `hashicorp/cap` library to support
Kubernetes 1.20 tokens.

- PR #148 
- PR #149 (old version of this PR)
- PR https://github.com/hashicorp/cap/pull/50

# Contributor Checklist
- [x] Backwards compatible


path_login_test.go
```diff
@@ -34,9 +40,188 @@ var (
 	testProjectedUID         = "77c81ad7-1bea-4d94-9ca5-f5d7f3632331"
 	testProjectedMockFactory = mockTokenReviewFactory(testProjectedName, testNamespace, testProjectedUID)
 
-	testDefaultPEMs = []string{testECCert, testRSACert}
+	// generate with: openssl ecparam -genkey -name prime256v1
+	ecdsaPrivateKeyText = `-----BEGIN EC PRIVATE KEY-----
```
tomhjp, line 44: Annoyingly, test data like this can often trigger false positives in automated scanners. Could we generate the key material dynamically using Go's crypto libraries instead?

tomhjp (approved):
I love the test updates 👍 just one comment on generating the test data to avoid causing queries down the road

swenson:
Thanks!

PR #153, opened 2022-06-05
Add PATCH support for kubernetes/config

remilapeyre:
It can be usefull to be able to change the value of some parameters in
the config but not others.

For example when changing disable_iss_validation we needed to keep
token_reviewer_jwt as was already set, but since this value is not
returned by pathConfigRead() there is no way to do this if you do not
already do know what value it was set to.

This patch adds PATCH support to the kubernetes/config endpoint so that
any parameter can be updated while not changing the other ones.

### DIFF
--- a/caching_file_reader_test.go
+++ b/caching_file_reader_test.go
@@ -27,7 +27,7 @@ func TestCachingFileReader(t *testing.T) {
 		})
 
 	// Write initial content to file and check that we can read it.
-	ioutil.WriteFile(f.Name(), []byte(content1), 0644)
+	ioutil.WriteFile(f.Name(), []byte(content1), 0o644)
 	got, err := r.ReadFile()
 	if err != nil {
 		t.Error(err)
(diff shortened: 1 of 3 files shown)


swenson:
Thanks @remilapeyre. This looks great.

In general I like this change, but I know we're hesitant to accept a PATCH endpoint on an auth plugin, partially because it isn't supported by the CLI, but also because it might dictate the semantics of how we use PATCH on other auth plugins in the future.

For example, in the KV secrets engine, we *do* [support PATCH](https://www.vaultproject.io/api-docs/secret/kv/kv-v2#patch-secret), but we explicitly use [RFC 7386 JSON merge patch](https://datatracker.ietf.org/doc/html/rfc7386) with an explicit content type set so that the user know what semantics they are signing up for.

Moving forward, I think the Vault team does want to have some sort of PATCH support for auth plugins, but we want to be introduce it more carefully and widely, and have some specific guidance and documentation on it.

Sorry that we can't accept this right now!

swenson:
Thanks @remilapeyre. This looks great.

In general I like this change, but I know we're hesitant to accept a PATCH endpoint on an auth plugin, partially because it isn't supported by the CLI, but also because it might dictate the semantics of how we use PATCH on other auth plugins in the future.

For example, in the KV secrets engine, we *do* [support PATCH](https://www.vaultproject.io/api-docs/secret/kv/kv-v2#patch-secret), but we explicitly use [RFC 7386 JSON merge patch](https://datatracker.ietf.org/doc/html/rfc7386) with an explicit content type set so that the user know what semantics they are signing up for.

Moving forward, I think the Vault team does want to have some sort of PATCH support for auth plugins, but we want to be introduce it more carefully and widely, and have some specific guidance and documentation on it.

Sorry that we can't accept this right now!

PR #156, opened 2022-07-18
VAULT-6727 Role resolution for K8S Auth

VioletHynes:
Support role resolution functionality (to work with role based quotas). The intent of this operation is to take the same data as a login request and return the role that would be used in the request, so that we can quota upon login requests that have a certain role.

Afterwards, I will push a new tag and pull this new version into Vault.


path_config_test.go
```diff
@@ -536,6 +537,61 @@ func TestConfig_LocalJWTRenewal(t *testing.T) {
 	}
 }
 
+func TestResolveRole(t *testing.T) {
```
VioletHynes, line 540: Woops, this is in the wrong file. Let me fix it...

path_login_test.go
```diff
@@ -1474,3 +1475,58 @@ func requireErrorCode(t *testing.T, err error, expectedCode int) {
...
+		NumUses:                  12,
+		BoundCIDRs:               nil,
+		AliasNameSource:          aliasNameSourceDefault,
+	}
+
+	entry, err := logical.StorageEntryJSON("role/"+role, validRoleStorageEntry)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := storage.Put(context.TODO(), entry); err != nil {
```
tomhjp, line 1507: nit: functionally they're the same, but Background signals the intent better here

```suggestion
	if err := storage.Put(context.Background(), entry); err != nil {
```

path_login_test.go
```diff
@@ -1474,3 +1475,58 @@ func requireErrorCode(t *testing.T, err error, expectedCode int) {
...
+		t.Fatal(err)
+	}
+
+	loginData := map[string]interface{}{
```
tomhjp, line 1511: nit: There are a few assumptions in the code about how non-existent roles are handled, so it might be nice to add one test case where a non-existent role is resolved as well.

tomhjp (approved):
LGTM, just a couple of optional nits in the tests.

tomhjp (approved):
LGTM

Issue #159, opened 2022-09-01
JWT signing algorithm ES384 became unsupported since version 0.13.0


jeroenvdheuvel:
The JWT signing algorithm ES384 became unsupported after [switch from golang-jwt to cap+jose](https://github.com/hashicorp/vault-plugin-auth-kubernetes/commit/f706889f82e62f413c5465f59c4859a1ea8f6166#diff-e9cabdd28848f01142dd2db22533f3f7e7ffd87d064f2edbb7149f1b6d4bbecdL26). Likely due to a typo the RS384 algorithm is supported twice while ES384 is no longer supported.
![image](https://user-images.githubusercontent.com/5213151/187868813-888b2311-c14c-4936-a04d-71df10f359b0.png)

[Kubernetes officially supports this algorithm](https://github.com/kubernetes/kubernetes/blob/v1.25.0/pkg/serviceaccount/jwt.go#L152). That means that Kubernetes clusters that use this algorithm aren't able to upgrade to Vault 1.11 successfully. Because Vault 1.11.* is using version 1.13.* of the vault-plugin-auth-kubernetes.

hsimon-hashicorp:
Closing as the linked PR was merged. :)

PR #160, opened 2022-09-01
Add support for JWT signature algorithm ES384 (support removed since 1.13.0)

jeroenvdheuvel:
# Overview
Add support for the previously supported JWT signing algorithm ES384. Support for this algorithm is likely removed accidentally in the change [switch from golang-jwt to cap+jose](https://github.com/hashicorp/vault-plugin-auth-kubernetes/commit/f706889f82e62f413c5465f59c4859a1ea8f6166#diff-e9cabdd28848f01142dd2db22533f3f7e7ffd87d064f2edbb7149f1b6d4bbecdL26).
image

Kubernetes supports [this algorithm](https://github.com/kubernetes/kubernetes/blob/v1.25.0/pkg/serviceaccount/jwt.go#L152), without it some Kubernetes clusters can't authenticate to Vault.

[Issue #159](https://github.com/hashicorp/vault-plugin-auth-kubernetes/issues/159)

### DIFF
--- a/path_login.go
+++ b/path_login.go
@@ -23,7 +23,7 @@ var defaultJWTIssuer = "kubernetes/serviceaccount"
 // See https://datatracker.ietf.org/doc/html/rfc7518#section-3.
 var supportedJwtAlgs = []capjwt.Alg{
 	capjwt.RS256, capjwt.RS384, capjwt.RS512,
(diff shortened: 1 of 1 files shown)


tomhjp (approved):
Thanks for finding and fixing this, and sorry for the inconvenience. I'll get this fix backported into Vault 1.11.4, and it should also be included in 1.12.0.

PR #161, opened 2022-09-05
Backport 1.11.x: Add support for JWT signature algorithm ES384 (#160)

tomhjp:
The JWT signature algorithm wasn't supported, while RS384 was twice.
Without this algorithm some Kubernetes clusters can't authenticate.


swenson (approved):
LGTM

Issue #163, opened 2022-09-29
allow patching of roles


f4z3r:
Currently it is not possible to modify roles once created.

We are encountering a use-case where we need to modify the policies attached to the tokens created by a role. A workaround would be to read the role, delete it, and re-create a modified version of it. This is not very nice.

Ideally we would want to be able to send a patch request to modify some parameters from the role configuration, in order to avoid the deletion of the role.

It should be noted that this would not modify the policies of the currently outstanding tokens that were created from that role before it got patched.

FYI: @zkck

f4z3r:
Nevermind, we looked at the code and a patch seems to be actually supported. This is mostly a documentation issue then I guess.

maxcoulombe:
Hey, thanks for bringing this to our attention. I agree with you it should be more obvious that the `create` command or request can also be used to update an existing role.

Would it be enough to simply change `Create Role` to `Create or Update Role` in the [API doc page for the K8s auth plugin](https://developer.hashicorp.com/vault/api-docs/auth/kubernetes#create-role)? When you were looking through the existing doc s are there other places where you would have expected this behaviour to be explained?

f4z3r:
Hey, seems enough to me. I mostly use the API reference, I have not looked through documentation or tutorials in details. But flying over them, it seems there are no changes required there.

maxcoulombe:
Should help clarify the behaviour so people do not have to waste time going through the source code in the future:
https://github.com/hashicorp/vault/pull/17371

cipherboy:
@f4z3r Correct, when fields are omitted on the POST, they are treated as if the defaults had been specified again. This confused a lot of people, hence why PKI has gone and started adding proper PATCH support.

(E.g., you could set `allow_localhost=false` and update the role via a say, `vault write pki/roles/existing allowed_domains="example.com"` and Vault would reset it to the default `allow_localhost=true`).

f4z3r:
So what is your suggestion? The callback under which the current "update" is registered is `logical.UpdateOperation`, which is a POST I would guess. Preventing an update via POST and registering the function under a PATCH would not be backwards compatible though. What is your approach regarding such changes, considering there was no documentation about the update capabilities of the endpoint?

I could find some time in the coming days to fix the implementation to properly behave like a POST, and do something inline with the PKI secret engine to also have a PATCH with the current implementation if you guys want. Just let me know if such a change would be fine in terms of compatibility in your eyes.

[...]

PR #164, opened 2022-10-07
fix role update endpoint and provide patch method

f4z3r:
# Overview

Fix the update endpoint for the role configuration to behave according to POST RESTful definitions. Add a PATCH endpoint for the role configuration.

This change might affect users that misuse the POST endpoints to update individual fields in the role configuration. If they omit required fields, this will now return an error, and if they simply omit fields, these will not be changed back to defaults (as opposed to keeping the value that was previously configured).

# Design of Change

I reused most of the code that used to be the update function for the patch method (as it actually behaved like a patch). Also refactored some code into own methods to reduce code duplication.

[ ] [Issue #163](https://github.com/hashicorp/vault-plugin-auth-kubernetes/issues/163)
[ ] [Vault PR #17371](https://github.com/hashicorp/vault/pull/17371)

# Contributor Checklist

[My Docs PR Link](link)
[] Backwards compatible

### DIFF
--- a/path_role.go
+++ b/path_role.go
@@ -228,31 +228,7 @@ func (b *kubeAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical
 	}
 
 	// Handle upgrade cases
-	{
-		if err := tokenutil.UpgradeValue(data, "policies", "token_policies", &role.Policies, &role.TokenPolicies); err != nil {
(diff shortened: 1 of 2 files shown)


f4z3r:
Hi @tvoran , is there anything else that you would want done in this PR?

PR #176, opened 2022-12-20
CreateOperation should only be implemented alongside ExistenceCheck

maxb:
See hashicorp/vault#18492

### DIFF
--- a/path_config.go
+++ b/path_config.go
@@ -88,7 +88,6 @@ then this plugin will use kubernetes.io/serviceaccount as the default issuer.
 		},
 		Callbacks: map[logical.Operation]framework.OperationFunc{
 			logical.UpdateOperation: b.pathConfigWrite,
-			logical.CreateOperation: b.pathConfigWrite,
 			logical.ReadOperation:   b.pathConfigRead,
 		},
 
--- a/path_config_test.go
+++ b/path_config_test.go
@@ -51,7 +51,7 @@ func TestConfig_Read(t *testing.T) {
 	}
 
 	req := &logical.Request{
-		Operation: logical.CreateOperation,
+		Operation: logical.UpdateOperation,
 		Path:      configPath,
 		Storage:   storage,
 		Data:      data,
@@ -91,7 +91,7 @@ func TestConfig(t *testing.T) {
 	}
 
 	req := &logical.Request{
-		Operation: logical.CreateOperation,
+		Operation: logical.UpdateOperation,
 		Path:      configPath,
 		Storage:   storage,
 		Data:      data,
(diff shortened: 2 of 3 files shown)


benashz:
I suppose an alternative approach would be to add the proper `ExistenceCheck` function, and preserve the backend's current behavior.  I wonder if you had considered that approach?

maxb:
This PR **does** preserve the backend's current behaviour. Applying it will not change the behaviour of the API that users experience.

It is not clear to me what you would consider to be a proper `ExistenceCheck` for a `/config` endpoint. I'd consider that to be no `ExistenceCheck`, because from the user perspective, they "create" the backend when they enable/mount it, and then they "update" the configuration of it - whilst the first ever write of a backend's config following mounting will be a create from the perspective of Vault's logical storage layer, I don't think users see it that way.

Vault is not very consistent in this though - I did a quick assessment of the capabilities supported by existing /config endpoints in Vault:

update only:

- /auth/cert/config
- /auth/gcp/config
- /auth/github/config
- /auth/jwt/config
- /auth/kerberos/config
- /auth/kubernetes/config
- /auth/ldap/config
- /gcp/config
- /identity/oidc/config
- /kv-v2/config
- /mongodbatlas/config

update and delete:

- /auth/cf/config
- /ad/config
- /alicloud/config
- /kubernetes/config

create and update:

- /auth/centrify/config
- /auth/okta/config
- /auth/radius/config

create, update and delete:

- /auth/azure/config
- /auth/oci/config
- /azure/config
- /gcpkms/config
- /openldap/config
- /terraform/config

benashz:
> This PR **does** preserve the backend's current behaviour. Applying it will not change the behaviour of the API that users experience.

I was more thinking of the behavior expected by the user, in the case of the create flow w.r.t ACL policies.

> 
> It is not clear to me what you would consider to be a proper `ExistenceCheck` for a `/config` endpoint. I'd consider that to be no `ExistenceCheck`, because from the user perspective, they "create" the backend when they enable/mount it, and then they "update" the configuration of it - whilst the first ever write of a backend's config following mounting will be a create from the perspective of Vault's logical storage layer, I don't think users see it that way.

No configuration exists in the storage is the case I was thinking of.

> 
> Vault is not very consistent in this though - I did a quick assessment of the capabilities supported by existing /config endpoints in Vault:

Definitely true. It would be ideal if there was one common model being followed, but that is not always possible.

maxb:
> I was more thinking of the behavior expected by the user, in the case of the create flow w.r.t ACL policies.

Yes, this is preserved unchanged.

benashz (approved):
LGTM!

Issue #175, opened 2022-12-14
The audience claim is not checked, the parameter `audience` in the Kubernetes auth role is ineffective


maelvls:
Vault 1.12 (and all previous versions) has a bug that prevents checking the JWT audience. If you configure a Kubernetes auth role with the parameter [`audience`](https://developer.hashicorp.com/vault/api-docs/auth/kubernetes) set to `vault` and expect Vault to refuse JWTs that don't have this audience, Vault will still accept any audience:

```bash
vault write auth/kubernetes/role/cert-manager \
    bound_service_account_names=vault \
    bound_service_account_namespaces=cert-manager \
    audience=vault   # <-- ignored 🔥
```

The Vault documentation [indicates](https://developer.hashicorp.com/vault/api-docs/auth/kubernetes) that the audience claim will be verified:

> `audience` (string: "") – _Optional_ Audience claim to verify in the JWT.

The code in [path_login.go](https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/eabe60240605c4cc3c2d73038931c2fbf47ff6aa/path_login.go#L408) goes something like this:

```go
parsedClaims := serviceAccount{}
_ = mapstructure.Decode(jwtClaims, &parsedClaims)
r, _ := tr.Review(ctx, client, jwtStr, parsedClaims.Audience)
```

Vault is checking that the audience in the JWT matches the audience in the JWT, instead of checking that the audience in the JWT matches the role's audience.

**Context:** as part of [The Vault issuer can now be given a serviceAccountRef (PR 5502)](https://github.com/jetstack/cert-manager/pull/5502), we wanted to use the audience claim verification to prevent someone hijacking a service account by creating a cert-manager Issuer. 

## Reproducing

First, setup Kubernetes 1.25 with Vault 1.12.1:

```shell
kind create cluster --image kindest/node:v1.25.3
helm upgrade --install vault hashicorp/vault --version 0.23.0 --set server.dev.enabled=true --set server.logLevel=debug --set global.tlsDisable=true --wait
kubectl exec vault-0 -i -- vault auth enable kubernetes
kubectl exec vault-0 -i -- sh -c 'vault write auth/kubernetes/config \
   token_reviewer_jwt="$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
   kubernetes_host="https://$KUBERNETES_PORT_443_TCP_ADDR:443" \
[...]

PR #179, opened 2023-01-20
Update audience handling for correctness

tomhjp:
I took a deeper look at our audience validation logic in response to #175, and saw a few things that could be improved, using the [TokenReview API docs](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-review-v1/) as a reference. I still don't think the logic is 100% correct when populating the TokenReview request audiences, but I think we have made some slightly incompatible promises in the Vault API, so I'm hesitant to align completely with the intended usage.

* Return 403 instead of 500 when audience validation fails
* Pass our expected audience instead of the token's audience to TokenReview
* Validate the TokenReview server is audience-aware by checking the returned audiences

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Changes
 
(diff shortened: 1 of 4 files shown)


swenson (approved):
LGTM; makes sense to me.

Issue #57, opened 2019-03-08
Don't mandate that the "token_reviewer_jwt" authentication token is a JWT


seh:
Given that this plugin takes the value of the "token_reviewer_jwt" configuration parameter and [stuffs it into an "Authorization" HTTP header as a bearer token](https://sourcegraph.com/github.com/hashicorp/vault-plugin-auth-kubernetes@db96aa4ab438cbc1cf544cec758d0d16ca4e9681/-/blob/token_review.go#L82-91), why does [the plugin care at all about the format of this token](https://sourcegraph.com/github.com/hashicorp/vault-plugin-auth-kubernetes@db96aa4ab438cbc1cf544cec758d0d16ca4e9681/-/blob/path_config.go#L96-103)?

For clusters with custom Webhook token authentication, tokens can be in any format: they can be service account tokens, or JWTs, [PASETOs](https://tools.ietf.org/id/draft-paragon-paseto-rfc-00.html), or something more exotic understood by the Webhook. The API server's authentication subsystem will take any value submitted as a bearer token and pass it through all the active token-consuming authentication methods. In our clusters, we have a Webhook that can consume tokens that are not JWTs. At present, we can't allow this plugin to authenticate using such a token.

I understand that it's not tenable to change the name of the "token_reviewer_jwt" configuration parameter today, but can we at least consider relaxing the enforcement of the format? If the intended consumer of the tokens doesn't mandate a particular format, why should this plugin be an overly stringent intermediary? In the interest of trying to help catch mistakes early, it's actually cutting off otherwise valid input.

This request may overlap with the earlier #34.

seh:
Ah, reading the code further, I now see that this plugin expects the token to authenticate as a Kubernetes service account, as [it destructures the user name to glean the account name and containing namespace](https://sourcegraph.com/github.com/hashicorp/vault-plugin-auth-kubernetes@db96aa4ab438cbc1cf544cec758d0d16ca4e9681/-/blob/token_review.go#L120-135). Given that, still, it matters less that the input is a JWT; what's apparently important is that it maps to a user name that looks like a service account's fully-qualified name.

yesteph:
Hi @seh 

[...]

PR #188, opened 2023-03-14
Remove obsolete tokenReviewer validation

abinet:
This change addresses https://github.com/hashicorp/vault-plugin-auth-kubernetes/issues/57

Removing the validation will allow to use non-JWT tokens for TokenReview call and cover the integration cases with Rancher, Webhook token authentication etc.

### DIFF
--- a/path_config.go
+++ b/path_config.go
@@ -14,7 +14,6 @@ import (
 
 	"github.com/hashicorp/vault/sdk/framework"
 	"github.com/hashicorp/vault/sdk/logical"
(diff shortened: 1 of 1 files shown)


esen-transporeon:
Are there any updates? We would also need this feature.

tvoran (commented):
Thanks for the patch! I think this will be workable, but would you mind updating the parameter description on line 44 to mention that it's a JWT or other token type?

tvoran:
Continued in #207

PR #200, opened 2023-07-04
Use port forwarding in integration tests

tomhjp:
Switch to port forwarding in the integration tests for communicating between the go tests and Vault within the Kubernetes cluster.

Eliminates some of the previous unsightly requirements:
* `kubectl` and exec'ing it
* Patch for kind config
* Patch for Vault helm chart manifest
* Specific host port 38200 being available to bring the kind cluster up

To avoid importing k8s client-go into the main plugin package, I gave the integration test its own go module, which generated a lot of lines of diff, but the rest of the changes are not too big.


Makefile
```diff
@@ -51,9 +49,10 @@ vault-image:
 # Create Vault inside the cluster with a locally-built version of kubernetes auth.
 .PHONY: setup-integration-test
 setup-integration-test: teardown-integration-test vault-image
-	kind --name ${KIND_CLUSTER_NAME} load docker-image hashicorp/vault:dev
-	kubectl create namespace test
-	helm install vault vault --repo https://helm.releases.hashicorp.com --version=0.23.0 \
+	kind --name $(KIND_CLUSTER_NAME) load docker-image hashicorp/vault:dev
+	kubectl --context="kind-$(KIND_CLUSTER_NAME)" create namespace test
+	helm upgrade --install vault vault --repo https://helm.releases.hashicorp.com --version=0.25.0 \
```
kschoche, line 54: I had to do a bit of digging on this change, it looks like we're calling `setup-integration-test` for each run of the test matrix which originally would do a `helm install`, since the test matrix doesnt appear to have changed at all did this test always fail in the past?
..just trying to wrap my head around why this change was made, not disagreeing with the approach :)

kschoche (approved):
Looks good to me!

Makefile
tomhjp: Sorry this is actually an unnecessary change - it worked fine before too. I recently learned about `upgrade --install` and prefer it over `install` because it's idempotent. The `--install` command just instructs helm to install the chart if it's not already installed, which means you can use the same command for installs and upgrades. It doesn't change anything in this case though because we have the teardown target as a dep of this target, so we're always going to be installing into a freshly cleaned up cluster anyway.

tomhjp:
Thanks!

PR #205, opened 2023-08-30
Automated dependency upgrades

hc-github-team-secure-vault-ecosystem:
Full log: https://github.com/hashicorp/vault-plugin-auth-kubernetes/actions/runs/6030090725

PR #206, opened 2023-08-30
update dependencies

raymonstah:


go.work
```diff
@@ -1,6 +0,0 @@
-go 1.20
```
raymonstah, line 1: @tomhjp FYI

According to https://go.googlesource.com/proposal/+/master/design/45713-workspace.md
> These go.work files should not be checked into the repositories so that they don‘t override the workspaces users explicitly define. Checking in go.work files could also lead to CI/CD systems not testing the actual set of version requirements on a module and that version requirements among the repository’s modules are properly incremented to use changes in the modules.

Safe to remove? It caused some weirdness in our automated dependency update workflow. See PR here: https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/205

vinay-gopalan (approved):

go.work
tomhjp: Oops, TIL. Thanks for the link! Yeah it's safe to remove, I only added it as a convenience to make `gopls` able to understand multiple modules.

tomhjp (approved):

fairclothjm (approved):
LGTM! May as well add a changelog entry as well

PR #207, opened 2023-09-07
Allow any token type for TokenReviewer validation

### LINKED ISSUE #57: Don't mandate that the "token_reviewer_jwt" authentication token is a JWT

tvoran:
# Overview

Continues #188, removing the requirement that `token_reviewer_jwt` be a JWT token, to support platforms like Rancher which use a different token type for K8s API auth.

- closes #57
- #188

# Contributor Checklist

- [x] docs update: https://github.com/hashicorp/vault/pull/22857
- [x] Backwards compatible


tomhjp (approved):
LGTM!

kschoche (approved):
🤘

tvoran:
Thanks!

PR #182, opened 2023-02-05
improvement: allow namespace selection in kubernetes authentication to go over label selection

f4z3r:
# Overview

Add possibility to reference the namespace from which ServiceAccounts authenticate via a label selector. This is useful in scenarios where Kubernetes namespaces are provided to teams by infrastructure teams, and standardized service accounts should be enabled to perform actions against Vault. At the moment, adding a namespace and service account within that namespace implies modifying the role on the Kubernetes authentication method. With this improvement, infrastructure teams can control what namespaces are allowed to connect to Vault via labels on the namespace itself.

# Design of Change

See discussion in https://github.com/hashicorp/vault/issues/16222.

# Related Issues/Pull Requests

[ ] [Issue #155](https://github.com/hashicorp/vault-plugin-auth-kubernetes/issues/155)

# Contributor Checklist

[x] Add relevant docs to upstream Vault repository, or sufficient reasoning why docs won’t be added yet
hashicorp/vault#19318
[x] Backwards compatible

Issue #155, opened 2022-07-13
Allow specifying a namespace selector for the allowed namespace from which to authenticate


f4z3r:
See issue hashicorp/vault#16222

We would like to replicate the change of the Secret Engine to this Auth engine, and allow specifying a label selector to define from which namespaces clients are allowed to authenticate with their ServiceAccounts.

FYI: @raffaelespazzoli

PR #218, opened 2023-11-27
VAULT-6936 support bound service account namespace selector

thyton:
# Overview
A high level description of the contribution, including:
What is the change? Support a label selector to define from which namespaces clients are allowed to authenticate with their ServiceAccounts.
Why is the change needed? This will add flexibility to bound namespace specification. A similar feature, [allowed_kubernetes_namespace_selector](https://developer.hashicorp.com/vault/api-docs/secret/kubernetes#allowed_kubernetes_namespace_selector), was also seen in Secret Engine.
How does this change affect the user experience (if at all)? No

# Design of Change
How was this change implemented? The change is based on [PR #182](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/182) and how [allowed_kubernetes_namespace_selector](https://developer.hashicorp.com/vault/api-docs/secret/kubernetes#allowed_kubernetes_namespace_selector) was implemented.

[ ] [Issue #155](https://github.com/hashicorp/vault-plugin-auth-kubernetes/issues/155)
[ ] [PR #182](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/182)

# Contributor Checklist
[Docs PR Link](https://github.com/hashicorp/vault/pull/19318)
```
make integration-test
cd integrationtest && INTEGRATION_TESTS=true CGO_ENABLED=0 KUBE_CONTEXT="kind-vault-plugin-auth-kubernetes" go test '-test.v' -count=1 -timeout=20m ./...
?   	github.com/hashicorp/vault-plugin-auth-kubernetes/integrationtest/k8s	[no test files]
=== RUN   TestSuccess
--- PASS: TestSuccess (0.48s)
=== RUN   TestSuccessWithTokenReviewerJwt
--- PASS: TestSuccessWithTokenReviewerJwt (0.10s)
=== RUN   TestSuccessWithNamespaceLabels
--- PASS: TestSuccessWithNamespaceLabels (0.11s)
=== RUN   TestFailWithBadTokenReviewerJwt
--- PASS: TestFailWithBadTokenReviewerJwt (0.11s)
=== RUN   TestUnauthorizedServiceAccountErrorCode
--- PASS: TestUnauthorizedServiceAccountErrorCode (0.10s)
=== RUN   TestAudienceValidation
=== RUN   TestAudienceValidation/config:_default,_JWT:_default
=== RUN   TestAudienceValidation/config:_default,_JWT:_a
=== RUN   TestAudienceValidation/config:_a,_JWT:_a
=== RUN   TestAudienceValidation/config:_a,_JWT:_b
=== RUN   TestAudienceValidation/config:_unset,_JWT:_default
=== RUN   TestAudienceValidation/config:_unset,_JWT:_a
--- PASS: TestAudienceValidation (0.61s)
    --- PASS: TestAudienceValidation/config:_default,_JWT:_default (0.10s)
    --- PASS: TestAudienceValidation/config:_default,_JWT:_a (0.09s)
    --- PASS: TestAudienceValidation/config:_a,_JWT:_a (0.11s)
    --- PASS: TestAudienceValidation/config:_a,_JWT:_b (0.09s)
    --- PASS: TestAudienceValidation/config:_unset,_JWT:_default (0.09s)
    --- PASS: TestAudienceValidation/config:_unset,_JWT:_a (0.10s)
PASS
ok  	github.com/hashicorp/vault-plugin-auth-kubernetes/integrationtest	2.077s
```
[ ] Backwards compatible


namespace_validate.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func namespaceValidateAPIFactory(config *kubeConfig) namespaceValidator {
+	return &namespaceValidateAPI{
+		config: config,
+	}
+}
+
+func (v *namespaceValidateAPI) ValidateLabels(ctx context.Context, client *http.Client, namespace string, namespaceSelector string) (bool, error) {
+	labelSelector, err := makeLabelSelector(namespaceSelector)
```
benashz, line 37: It doesn't look like we are guarding against empty namespace, nor empty namespaceSelector. I think we might want to return an error if either one is empty.

namespace_validate.go
```diff
@@ -0,0 +1,118 @@
...
+	req, err := http.NewRequestWithContext(ctx, "GET", url, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// Use the configured TokenReviewer JWT as the bearer
+	if v.config.TokenReviewerJWT == "" {
+		return nil, errors.New("namespace lookup failed: TokenReviewer JWT needs to be configured to use namespace selectors")
+	}
+	bearer := fmt.Sprintf("Bearer %s", v.config.TokenReviewerJWT)
```
benashz, line 73: nit: `bearer` is only used once. Might want to consider in-lining it `setRequestHeader()` below.

namespace_validate.go
```diff
@@ -0,0 +1,118 @@
...
+	labelSelector := metav1.LabelSelector{}
+	decoder := k8s_yaml.NewYAMLOrJSONDecoder(strings.NewReader(selector), len(selector))
+	err := decoder.Decode(&labelSelector)
+	if err != nil {
+		return labelSelector, err
+	}
+	return labelSelector, nil
+}
+
+func (v *namespaceValidateAPI) getNamespaceLabels(ctx context.Context, client *http.Client, namespace string) (map[string]string, error) {
```
benashz, line 62: I wonder if it make things simpler if we starting the k8s client instead of this homebrew client. We should consider that for a future release.

namespace_validate.go
```diff
@@ -0,0 +1,118 @@
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != 200 {
+		return nil, fmt.Errorf("failed to get namespace (code %d): %s", resp.StatusCode, body)
```
benashz, line 85: I would, caution against including the body in the error since it may or may not contain confidential info. I think there is way to extract the error message from using some k8s client API code. Might be worth considering.

namespace_validate.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	ns := v1.Namespace{}
+
```
benashz, line 88: gratuitous whitespace
```suggestion
```

namespace_validate.go
```diff
@@ -0,0 +1,118 @@
...
+		return nil, err
+	}
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != 200 {
+		return nil, fmt.Errorf("failed to get namespace (code %d): %s", resp.StatusCode, body)
+	}
+	ns := v1.Namespace{}
```
benashz, line 87:
```suggestion
	var ns v1.Namespace
```

benashz (commented):
Gave it a first pass. I think we may want to revisit the approach a bit. I had some questions around request paramater types JSON+YAML, the way that namespaces and namespace selector's are evaluated together.

Overall, it seems like the change is heading in a positive direction.

path_role.go
```diff
@@ -49,6 +49,12 @@ are allowed.`,
 					Type: framework.TypeCommaStringSlice,
 					Description: `List of namespaces allowed to access this role. If set to "*" all namespaces
 are allowed.`,
+				},
+				"bound_service_account_namespace_selector": {
+					Type: framework.TypeString,
+					Description: `A label selector for Kubernetes namespaces which are allowed to access this role.
+Accepts either a JSON or YAML object. If set with bound_service_account_namespaces,
+the conditions are ORed.`,
```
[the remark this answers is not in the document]
tomhjp: I think the answer to the additional questions is it's consistent with the k8s secret engine. In retrospect, using OR rather than AND does seem a bit surprising, but I think we should probably maintain consistency for now.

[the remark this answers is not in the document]
benashz: > I think the answer to the additional questions is it's consistent with the k8s secret engine. In retrospect, using OR rather than AND does seem a bit surprising, but I think we should probably maintain consistency for now.

@tomhjp  Ah, so you are referring to https://developer.hashicorp.com/vault/api-docs/secret/kubernetes#allowed_kubernetes_namespace_selector? I guess that will make the UX a bit more consistent.

namespace_validate.go
```diff
@@ -0,0 +1,118 @@
...
+	config *kubeConfig
+}
+
+func namespaceValidateAPIFactory(config *kubeConfig) namespaceValidator {
+	return &namespaceValidateAPI{
+		config: config,
+	}
+}
+
+func (v *namespaceValidateAPI) ValidateLabels(ctx context.Context, client *http.Client, namespace string, namespaceSelector string) (bool, error) {
```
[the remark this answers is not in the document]
thyton: Thank you! Will fix this and the other similar function in the mock struct.

path_login.go
```diff
@@ -247,7 +247,13 @@ func (b *kubeAuthBackend) aliasLookahead(ctx context.Context, req *logical.Reque
 	}
 	// validation of the JWT against the provided role ensures alias look ahead requests
 	// are authentic.
-	sa, err := b.parseAndValidateJWT(jwtStr, role, config)
+	client, err := b.getHTTPClient()
+	if err != nil {
+		b.Logger().Error(`Failed to get the HTTP client`, "err", err)
```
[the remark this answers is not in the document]
thyton: I confirm `err` contains no confidential information. https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/5dfb046eb2504bd56664a9f891f573baf7ecbf32/backend.go#L389-L398

The `err` values are https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/5dfb046eb2504bd56664a9f891f573baf7ecbf32/backend.go#L64-L65

namespace_validate.go
thyton: I'm leaning towards not using k8s client based on the [above discussion](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/218#discussion_r1407723039) for now.

path_role.go
```diff
@@ -161,8 +167,15 @@ func (b *kubeAuthBackend) pathRoleRead(ctx context.Context, req *logical.Request
 
 	// Create a map of data to be returned
 	d := map[string]interface{}{
-		"bound_service_account_names":      role.ServiceAccountNames,
-		"bound_service_account_namespaces": role.ServiceAccountNamespaces,
+		"bound_service_account_names": role.ServiceAccountNames,
+	}
+
+	if len(role.ServiceAccountNamespaces) > 0 {
+		d["bound_service_account_namespaces"] = role.ServiceAccountNamespaces
```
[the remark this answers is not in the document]
thyton: Thank you. I overlooked this.

path_role.go
```diff
@@ -304,13 +317,23 @@ func (b *kubeAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical
...
 		role.ServiceAccountNamespaces = namespaces.([]string)
-	} else if req.Operation == logical.CreateOperation {
-		role.ServiceAccountNamespaces = data.Get("bound_service_account_namespaces").([]string)
 	}
-	// Verify namespaces is not empty
-	if len(role.ServiceAccountNamespaces) == 0 {
-		return logical.ErrorResponse("%q can not be empty", "bound_service_account_namespaces"), nil
+
+	if namespaceSelector, ok := data.GetOk("bound_service_account_namespace_selector"); ok {
+		role.ServiceAccountNamespaceSelector = namespaceSelector.(string)
```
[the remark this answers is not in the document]
thyton: I missed it. Thank you

namespace_validate.go
thyton: The guarding step happens in path_role.go where `bound_service_account_namespaces`  and `bound_service_account_namespace_selector` are set. With `bound_service_account_namespace_selector`, the  PR allows either but not both of these fields to be empty.
https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/ca554dd48b7c13bc95f8610ba4ba590765aa368e/path_role.go#L318-L321

`ValidateLabels(...)` is used in path_login.go where the request's token namespace meets the namespace condition. I think we don't need the guarding step here as  path_login.go only references those fields. https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/ca554dd48b7c13bc95f8610ba4ba590765aa368e/path_login.go#L342-L356

Let me know if I misunderstood your point anywhere or how I can improve the guarding step.

namespace_validate.go
thyton: I've parsed the error status from body. Let me know your thought. Thank you.

thyton:
@benashz @tomhjp I appreciate the detailed first pass! It's ready for the next one whenever you have a chance.

namespace_validator.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil, err
+	}
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		errStatus := &metav1.Status{}
+		err = json.Unmarshal(body, errStatus)
+		if err == nil && errStatus.Status != metav1.StatusSuccess {
```
benashz, line 78: nit: it's more a personal preference, but adding an empty newline between blockes of related code blocks makes it a bit easier to read

```
if err != nil {
    return err
}

var ...
```

path_login.go
```diff
@@ -282,7 +288,7 @@ func (keySet DontVerifySignature) VerifySignature(_ context.Context, token strin
 }
 
 // parseAndValidateJWT is used to parse, validate and lookup the JWT token.
-func (b *kubeAuthBackend) parseAndValidateJWT(jwtStr string, role *roleStorageEntry, config *kubeConfig) (*serviceAccount, error) {
+func (b *kubeAuthBackend) parseAndValidateJWT(ctx context.Context, client *http.Client, jwtStr string, role *roleStorageEntry, config *kubeConfig) (*serviceAccount, error) {
```
benashz, line 291: nit: this line is bit long, would you mind  folding it?

path_login.go
```diff
@@ -333,7 +339,17 @@ func (b *kubeAuthBackend) parseAndValidateJWT(jwtStr string, role *roleStorageEn
...
 
 	// verify the namespace is allowed
-	if len(role.ServiceAccountNamespaces) > 1 || role.ServiceAccountNamespaces[0] != "*" {
+	valid := false
+	if role.ServiceAccountNamespaceSelector != "" {
+		if valid, err = b.namespaceValidatorFactory(config).validateLabels(ctx,
+			client, sa.namespace(), role.ServiceAccountNamespaceSelector); err != nil {
+			return nil, err
+		}
+	}
```
benashz, line 348: nit: add newline below.

path_login.go
```diff
@@ -333,7 +339,17 @@ func (b *kubeAuthBackend) parseAndValidateJWT(jwtStr string, role *roleStorageEn
...
+	if !valid && len(role.ServiceAccountNamespaces) == 0 {
+		return nil, logical.CodedError(http.StatusForbidden, "namespace not authorized")
+	}
+	if !valid && (len(role.ServiceAccountNamespaces) > 1 || role.ServiceAccountNamespaces[0] != "*") {
```
benashz, line 352: I am bit confused by these `if !valid` conditionals, I wonder if they belong under the conditional at line 343, since valid will always be false if `role.ServiceAccount...Selector` is empty string? I am guessing  valid only makes sense in the context of namespace label validation?

path_login.go
```diff
@@ -333,7 +339,17 @@ func (b *kubeAuthBackend) parseAndValidateJWT(jwtStr string, role *roleStorageEn
 	}
 
 	// verify the namespace is allowed
-	if len(role.ServiceAccountNamespaces) > 1 || role.ServiceAccountNamespaces[0] != "*" {
+	valid := false
```
benashz, line 342: I think you can make this `var valid bool`, since that will be false by default.

path_login_test.go
```diff
@@ -758,6 +774,35 @@ func TestLoginSvcAcctAndNamespaceSplats(t *testing.T) {
...
+		Data:      data,
+		Connection: &logical.Connection{
+			RemoteAddr: "127.0.0.1",
+		},
+	}
+
+	resp, err := b.HandleRequest(context.Background(), req)
+	if err != nil || (resp != nil && resp.IsError()) {
+		t.Fatalf("err:%s resp:%#v\n", err, resp)
+	}
```
benashz, line 803: Thanks for the tests! 

It looks like we never evaluate `resp`. It's not totally clear what the test is testing. I think maybe checking some on the resp would make the test a bit clearer, or adding a comment, or adding something to the name of the test.

path_login_test.go
```diff
@@ -758,6 +774,35 @@ func TestLoginSvcAcctAndNamespaceSplats(t *testing.T) {
 	}
 }
 
+func TestLoginSvcAcctNamespaceSelector(t *testing.T) {
```
benashz, line 777: I don't see the test for failed login, I wonder if we could turn this into a table driven testcase that tests both cases.

path_role.go
```diff
@@ -161,8 +167,9 @@ func (b *kubeAuthBackend) pathRoleRead(ctx context.Context, req *logical.Request
 
 	// Create a map of data to be returned
 	d := map[string]interface{}{
-		"bound_service_account_names":      role.ServiceAccountNames,
-		"bound_service_account_namespaces": role.ServiceAccountNamespaces,
+		"bound_service_account_names":              role.ServiceAccountNames,
+		"bound_service_account_namespaces":         role.ServiceAccountNamespaces,
+		"bound_service_account_namespace_selector": role.ServiceAccountNamespaceSelector,
```
benashz, line 172: :+1:

path_role.go
```diff
@@ -304,13 +311,22 @@ func (b *kubeAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical
...
-		role.ServiceAccountNamespaces = data.Get("bound_service_account_namespaces").([]string)
 	}
-	// Verify namespaces is not empty
-	if len(role.ServiceAccountNamespaces) == 0 {
-		return logical.ErrorResponse("%q can not be empty", "bound_service_account_namespaces"), nil
+
+	role.ServiceAccountNamespaceSelector = data.Get("bound_service_account_namespace_selector").(string)
+
+	// Verify namespaces is not empty unless selector is set
+	if len(role.ServiceAccountNamespaces) == 0 && role.ServiceAccountNamespaceSelector == "" {
```
benashz, line 319: nit: I see that we take the length of `role.ServiceAccountNamespaces` in multiple places, would be nice to factor that out to variable.

path_role.go
```diff
@@ -304,13 +311,22 @@ func (b *kubeAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical
...
+		return logical.ErrorResponse("%q can not be empty if %q is not set", "bound_service_account_namespaces", "bound_service_account_namespace_selector"), nil
```
benashz, line 320: nit: this line is a bit long, would you folding it around column 100?

path_role.go
```diff
@@ -364,6 +380,10 @@ type roleStorageEntry struct {
 	// role.
 	ServiceAccountNamespaces []string `json:"bound_service_account_namespaces" mapstructure:"bound_service_account_namespaces" structs:"bound_service_account_namespaces"`
 
+	// ServiceAccountNamespaceSelector is the label selector string of the
+	// namespaces able to access this role.
+	ServiceAccountNamespaceSelector string `json:"bound_service_account_namespace_selector" mapstructure:"bound_service_account_namespace_selector" structs:"bound_service_account_namespace_selector"`
```
benashz, line 385: Are we still using `structs`, and  `mapstructure` we can probably clean that up in a subsequent PR.

benashz (commented):
Thanks for making the requested changes! I think we are almost there.

integrationtest/integration_test.go
```diff
@@ -149,6 +157,23 @@ func TestSuccessWithTokenReviewerJwt(t *testing.T) {
 	}
 }
 
+func TestSuccessWithNamespaceLabels(t *testing.T) {
```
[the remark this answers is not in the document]
thyton: I will add it. Thank you

path_login.go
thyton: I will restructure this in a more readable way.

path_role.go
thyton: I confirm we are still using `structs` and `mapstructure` in other places.

path_login.go
```diff
@@ -333,12 +340,28 @@ func (b *kubeAuthBackend) parseAndValidateJWT(jwtStr string, role *roleStorageEn
 	}
 
 	// verify the namespace is allowed
-	if len(role.ServiceAccountNamespaces) > 1 || role.ServiceAccountNamespaces[0] != "*" {
-		if !strutil.StrListContainsGlob(role.ServiceAccountNamespaces, sa.namespace()) {
-			return nil, logical.CodedError(http.StatusForbidden, "namespace not authorized")
+	var allowed bool
+	codedErr := logical.CodedError(http.StatusForbidden, "namespace not authorized")
+	if len(role.ServiceAccountNamespaces) != 0 {
```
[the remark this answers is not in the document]
thyton: @benashz I wonder if explaining my thinking/reasoning could help clarify this namespace verify code. 

#### Prior to this PR:
https://developer.hashicorp.com/vault/api-docs/auth/kubernetes#bound_service_account_namespaces used to be a required field. 
- https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/5dfb046eb2504bd56664a9f891f573baf7ecbf32/path_role.go#L310-L313 makes sure len(role.ServiceAccountNamespaces) != 0 when a role is set. 
- https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/5dfb046eb2504bd56664a9f891f573baf7ecbf32/path_login.go#L335-L340 invalidates when the token SA namespace is not included in the explicitly set namespaces.



#### In this PR:
https://developer.hashicorp.com/vault/api-docs/auth/kubernetes#bound_service_account_namespaces is no longer a required field. Either but not both this and the new field `bound_service_account_namespace_selector` can be empty.

I have a different draft of code that evaluates explicitly set namespaces. This draft keeps some of the old checking logic `len(role.ServiceAccountNamespaces) > 1 || role.ServiceAccountNamespaces[0] != "*" ` seen prior to this PR, which might aid the transitioning from the old to new during code reviews.
```
// verify the namespace is allowed
allowed := true
codedErr := logical.CodedError(http.StatusForbidden, "namespace not authorized")
if len(role.ServiceAccountNamespaces) == 0 {
	allowed = false
} else if len(role.ServiceAccountNamespaces) > 1 || role.ServiceAccountNamespaces[0] != "*" {
	if !strutil.StrListContainsGlob(role.ServiceAccountNamespaces, sa.namespace()) {
		allowed = false
	}
}
```

My current code attempts to shorten/refactor the above draft by defaulting `allowed` to false, negating the if conditions, and setting `allowed` to `true` when met.
```
var allowed bool
codedErr := logical.CodedError(http.StatusForbidden, "namespace not authorized")
if len(role.ServiceAccountNamespaces) != 0 {
	if role.ServiceAccountNamespaces[0] == "*" ||
		strutil.StrListContainsGlob(role.ServiceAccountNamespaces, sa.namespace()) {
		allowed = true
	}
}

```

I don't have a very strong opinion on this. I can update to whichever version is best for the team/future developers :).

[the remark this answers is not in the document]
benashz: +1 for setting allowed to false by default.

It looks like you are only setting allowed to true whenever `role.ServiceAccountNamespaces[0] == "*"`. How do you handle the the case where other patterns are in `role.ServiceAccountNamespaces`? E.g  `["foo-*", "bar-"]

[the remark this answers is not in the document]
benashz: Ah, I see it now, was stuck on the extra folding at `||`.

benashz (commented):
Looking good! It would be good to add some tests for the methods provided by `namespaceValidatorWrapper{}`, as well as what we covered during our sync.

benashz (approved):
LGTM! Thanks for making the extra changes.

mbenson:
When are we likely to see a new release incorporating this change?

PR #132, opened 2022-02-02
Allow annotating service accounts with arbitrary tags for policy templating

dovys:
Addresses [hashicorp#85](https://github.com/hashicorp/vault-plugin-auth-kubernetes/issues/85)

We took Mark's proposed design and added support for defining annotations on service accounts that can later on be used in policy templating.

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: example-account
  namespace: default
  annotations:
    auth-metadata.vault.hashicorp.com/service-role: example-value
```

Which you can then use in policies as so
```
{{identity.entity.aliases.${vault_auth_backend.kubernetes.accessor}.metadata.service_role}}
```

**Who the change affects or is for (stakeholders)?**
The change is behind a config flag called `enable_custom_metadata_from_annotations` so it should not affect any of the existing integrations.
In order to enable the flag users will have to update the clusterrole and allow Vault to read service accounts. If this change is accepted we'll open PRs to update various docs, terraform providers, etc.
This will also introduce another roundtrip to the Kubernetes API, however we are using a pooled tcp client so hopefully not too many new open connections.

**Why is the change needed?**
At @monzo we want to use fully qualified service names for certificates issued under Vault's PKI, ie `service.profile`. At the same time we want policies and roles to be very strict so `service.profile` is not allowed to obtain certificates for `service.id`. As we have over 2000 microservices and they grow at a steady rate it's not feasible to write a new policy for each. Allowing us to annotate service accounts with the fully qualified service name and then use it in PKI mount allowed_domains with templating solves this issue.

# Design of Change
We've largely adhered to the current code style present in the repo. 
All changes are behind a config flag `enable_custom_metadata_from_annotations`. Existing integrations are not affected until the flag is turned on.

# Related Issues/Pull Requests
**TBD** - we'll create PRs for the terraform provider and docs when this PR has been largely agreed upon

# Contributor Checklist
- [ ] **TBD** Add relevant docs to upstream Vault repository, or sufficient reasoning why docs won’t be added yet
[My Docs PR Link](link)
[Example](https://github.com/hashicorp/vault/commit/2715f5cec982aabc7b7a6ae878c547f6f475bba6)
- [x] Add output for any tests not ran in CI to the PR description (eg, acceptance tests)
- [x] Backwards compatible

Issue #85, opened 2020-03-26
Add custom metadata to Vault from ServiceAccount annotations


mrknmc:
We would like to create policy rules that use custom entity/alias metadata and have this metadata be supplied from Kubernetes annotations on the ServiceAccount.

The way I envision this working is, if you have a ServiceAccount with annotations such as:

```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: example-account
  namespace: default
  annotations:
    # could have different form
    vault.hashicorp.com/auth-metadata/metadata-key: example-value
```

Then the auth plugin adds following metadata to the entity/alias:

```
metadata-key: example-value
```

We could then create a templated policy document (in Terraform) such as:

```
data "vault_policy_document" "tokens_write_document" {
  path         = "root/{{identity.entity.aliases.<mount-accessor>.metadata.<metadata-key>}}/*"
  capabilities = ["write", "list", "read"]
}
```

And with this policy in place, the ServiceAccount would have access to any secret under:

```
/root/example-value/*
```

Currently, we would have to manually edit entity/alias metadata to achieve this at which point we might as well create per-service-account policies which we would really like to avoid doing.

What are your thoughts on this? I would gladly submit a PR.

I'm thinking what would be necessary is to make an extra request to get the annotations for a given service account, parse them, and pass them to Vault in this code block: https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/master/path_login.go#L112-L133. Extra RBAC permissions would be required for the JWT reviewer service account too.

sfynx:
Yeah, I would very much appreaciate to have this feature, it sounds like a logical thing to have actually seeing how you can annotate everything in Kubernetes anyway. This would enable us to organize our secrets and their access control based on an existing k8s structure.

In my case I would like to define a secret as `root/<cluster name>/<application name>/<application environment>/somesecret`, so that these paths are human-readable for people to access. Right now I would have to abuse the existing namespace or service account name / token fields for that or do it manually like you said.

aarmstrong-figure:
I also have a use case for this feature as we're using the Kubernetes auth method and would like to take advantage of templated policies being defined ahead of time using a structured path that would require more metadata than just `service_account_name` and `service_account_namespace`.

sfynx:
Would still want this. Right now I'm creating a role and policy for each end user that needs to access part of the path, but it is a pain to manage.

mooneye14:
[...]

PR #226, opened 2024-01-02
set auth alias custom metadata to service account annotations

thyton:
# Overview
This PR sets auth alias custom metadata to service account annotations. Users can create policy rules that use custom entity/alias metadata and have this metadata be supplied from Kubernetes annotations on the ServiceAccount.

# Design of Change
Extra permission configuration for installations is required to allow to get annotations from Kubernetes. A new `use_annotations_as_alias_metadata` config option is added for users to explicitly opt-in to ensure existing installations still work if this option is opted out.

Only annotations with the prefix `vault.hashicorp.com/alias-metadata-` of the client token's
associated service account will be added to the auth alias's metadata upon a login request. For example, if an annotation "vault.hashicorp.com/alias-metadata-foo" is configured, "foo" with its value will be added.

NOTE that Vault will need permission to read service accounts from the Kubernetes API.

Example:
```
# Service Account "default" in "default" namespace
kubectl get serviceaccount default -n default -o json
{
    "apiVersion": "v1",
    "kind": "ServiceAccount",
    "metadata": {
        "annotations": {
            "auth-metadata.vault.hashicorp.com/alias-metadata-bar": "baz",
            "auth-metadata.vault.hashicorp.com/alias-metadata-foo": "bar"
        },
        "creationTimestamp": "2024-01-07T23:27:24Z",
        "name": "default",
        "namespace": "default",
        "resourceVersion": "5514",
        "uid": "ddae5be7-3aaa-4256-adc4-b77ea508e25e"
    }
}


# Configure a role "demo" with bound_service_account_names=default bound_service_account_namespaces=default
vault write auth/kubernetes/role/demo bound_service_account_names=default bound_service_account_namespaces=default policies=default ttl=1h
Success! Data written to: auth/kubernetes/role/demo


# Login request to vault.
vault write auth/kubernetes/login role=demo jwt=$(kubectl create token -n default default)
Key                                       Value
---                                       -----
token                                     hvs.CAESIAD96sjSFtZU6BXo5E1SuPoXcb_M_PJ9CSs0fM4OEcVMGh4KHGh2cy5kdTJid1JUeDRPZ3dXOVYzUTQxS3Zkc1Q
token_accessor                            FNQlfHrY2ioNmsKiX4EqACjJ
token_duration                            1h
token_renewable                           true
token_policies                            ["default"]
identity_policies                         []
policies                                  ["default"]
token_meta_role                           demo
token_meta_service_account_name           default
token_meta_service_account_namespace      default
token_meta_service_account_secret_name    n/a
token_meta_service_account_uid            ddae5be7-3aaa-4256-adc4-b77ea508e25e


# A new entity alias was created
vault list identity/entity-alias/id 
Keys
----
c474b739-f5d7-9b42-a8ea-d192972d832d

# The entity alias is expected to have the customized key value pairs from the SA annotations together with other pairs for vault internal use.
vault read identity/entity-alias/id/c474b739-f5d7-9b42-a8ea-d192972d832d
Key                          Value
---                          -----
canonical_id                 7281fda9-12c3-66aa-43d1-8e9817731d26
creation_time                2024-01-08T00:47:57.77138992Z
custom_metadata              <nil>
id                           c474b739-f5d7-9b42-a8ea-d192972d832d
last_update_time             2024-01-08T00:47:57.77138992Z
local                        false
merged_from_canonical_ids    <nil>
metadata                     map[foo:bar bar:baz service_account_name:default service_account_namespace:default service_account_secret_name: service_account_uid:ddae5be7-3aaa-4256-adc4-b77ea508e25e]
mount_accessor               auth_kubernetes-dev_6c913879
mount_path                   auth/kubernetes/
mount_type                   kubernetes-dev
name                         ddae5be7-3aaa-4256-adc4-b77ea508e25e
namespace_id                 root
```
- [x] [Issue #85](https://github.com/hashicorp/vault-plugin-auth-kubernetes/issues/85)
- [x] [PR #132](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/132)
# Contributor Checklist
[My Docs PR vault #24941 Link](https://github.com/hashicorp/vault/pull/24941)
- [x] Backwards compatible


path_login.go
```diff
@@ -139,20 +139,25 @@ func (b *kubeAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
...
+	err = sa.lookup(ctx, client, jwtStr, role.Audience, b.reviewFactory(config))
 	if err != nil {
 		b.Logger().Debug(`login unauthorized`, "err", err)
 		return nil, logical.ErrPermissionDenied
 	}
 
-	uid, err := serviceAccount.uid()
+	annotations, err := b.serviceAccountGetterFactory(config).annotations(ctx, client, sa.Namespace, sa.Name)
+	if err != nil {
+		b.Logger().Debug("failed to get service account annotations", "err", err)
```
tomhjp, line 156: I don't think there's a great need to log the error here as we are returning it to the user - the main reason we log the above error is because we only return an opaque permission denied and otherwise throw the error away. But by getting to here, we know the request can be treated pretty much like an authenticated request and I think it's ok to return the real error to the user.

service_account_getter.go
```diff
@@ -0,0 +1,108 @@
...
+	url := fmt.Sprintf("%s/api/v1/namespaces/%s/serviceaccounts/%s",
+		strings.TrimSuffix(w.config.Host, "/"), namespace, serviceAccount)
+	req, err := http.NewRequestWithContext(ctx, "GET", url, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// Use the configured TokenReviewer JWT as the bearer
+	if w.config.TokenReviewerJWT == "" {
+		return nil, errors.New("service account lookup failed: TokenReviewer JWT needs to be configured to retrieve service accounts")
```
tomhjp, line 48: Is there any reason we can't use the client's own token like we do for other API requests? It's not my favourite feature, but I would favour consistency and shared code to achieve that if possible.

tomhjp (commented):
Looking good! Not a complete review, but I left a couple of minor WIP thoughts

path_login.go
thyton: Thank you for the detailed feedback.

service_account_getter.go
```diff
@@ -0,0 +1,114 @@
...
+	"net/http"
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+	kubeerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const annotationKeyPrefix = "auth-metadata.vault.hashicorp.com/"
```
tomhjp, line 17: There are a couple of other k8s integrations that use Vault-like URLs - please could we verify that this is consistent to some degree with Agent injector's annotations and VSO? They are serving different purposes, but I'm not certain how consistently we use vault.hashicorp.com for the domain.

integrationtest/integration_test.go
```diff
@@ -220,6 +243,84 @@ func TestFailWithBadTokenReviewerJwt(t *testing.T) {
...
 }
 
+func TestAuthAliasMetadataAssignment(t *testing.T) {
+	// annotate the service account
+	expMetadata := map[string]string{
+		"foo": "bar",
+		"bar": "baz",
+	}
+
+	annotationPrefix := "auth-metadata.vault.hashicorp.com/"
```
tomhjp, line 253: nit: Makes the intent a little clearer

```suggestion
	const annotationPrefix = "auth-metadata.vault.hashicorp.com/"
```

integrationtest/integration_test.go
```diff
@@ -220,6 +243,84 @@ func TestFailWithBadTokenReviewerJwt(t *testing.T) {
...
+	annotations := map[string]string{}
+	for k, v := range expMetadata {
+		annotations[annotationPrefix+k] = v
+	}
+	annotateServiceAccount(t, "vault", annotations)
+
+	client, cleanup := setupKubernetesAuth(t, "vault", nil, nil)
+	defer cleanup()
+
+	loginSecret, err := client.Logical().Write("auth/kubernetes/login", map[string]interface{}{
```
tomhjp, line 263: This test looks good. One further improvement we could make though (and I think it would probably also be a net simplification) is to replace the entity API enumeration code with a straightforward secret-retrieval call that relies on a templated policy attached to the token we get from this call.

tomhjp (commented):
Generally LGTM - I think there is one major issue with backwards compatibility, but the remaining comments are minor suggestions/questions

path_login.go
```diff
@@ -139,44 +153,50 @@ func (b *kubeAuthBackend) pathLogin(ctx context.Context, req *logical.Request, d
...
-	err = serviceAccount.lookup(ctx, client, jwtStr, role.Audience, b.reviewFactory(config))
-
+	err = sa.lookup(ctx, client, jwtStr, role.Audience, b.reviewFactory(config))
 	if err != nil {
 		b.Logger().Debug(`login unauthorized`, "err", err)
 		return nil, logical.ErrPermissionDenied
 	}
 
-	uid, err := serviceAccount.uid()
+	annotations, err := b.serviceAccountGetterFactory(config).annotations(ctx, client, jwtStr, sa.namespace(), sa.name())
```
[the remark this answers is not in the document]
thyton: Thank you for catching this! I'll add the config option for it.

service_account_getter.go
thyton: Thank you. I've updated the prefix to `vault.hashicorp.com/alias-metadata-`.

integrationtest/integration_test.go
thyton: Thank you. I'll change the flow to use templated policies.

path_config.go
```diff
@@ -88,6 +88,17 @@ then this plugin will use kubernetes.io/serviceaccount as the default issuer.
 					Name: "Disable use of local CA and service account JWT",
 				},
 			},
+			"use_annotations_as_alias_metadata": {
+				Type: framework.TypeBool,
+				Description: `Indicate annotations, with the "vault.hashicorp.com/alias-metadata-"
+prefix, of the request JWT service account will be added to the alias metadata.
+Note Vault client or the token reviewer service account needs to be configured
+with the permission to get service account annotations from Kubernetes API.`,
```
tomhjp, line 96:
```suggestion
				Description: `Use annotations from the client token's associated service account
as alias metadata for the Vault entity. Only annotations with the prefix
"vault.hashicorp.com/alias-metadata-" will be used. Note that Vault
will need permission to read service accounts from the Kubernetes API.`,
```

backend.go
```diff
@@ -115,6 +115,14 @@ type kubeAuthBackend struct {
...
 	// tlsMu provides the lock for synchronizing updates to the tlsConfig.
 	tlsMu sync.RWMutex
+
+	// useAnnotationsAsAliasMetadata indicates that annotations, with the
+	// "vault.hashicorp.com/alias-metadata-" prefix, of the request JWT
+	// service account will be added to the alias metadata.
+	// Note Vault client or the token reviewer service account needs
+	// to be configured with the permission to get service account
+	// annotations from Kubernetes API
+	useAnnotationsAsAliasMetadata bool
```
tomhjp, line 125: I think this should be a member of `kubeConfig` instead. The lifecycle of fields in `kubeAuthBackend` matches the lifecycle of the plugin process, but this value is tied to the most recent write to the config endpoint instead.

path_config.go
```diff
@@ -189,6 +201,7 @@ func (b *kubeAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Requ
 		return nil, err
 	}
 
+	b.useAnnotationsAsAliasMetadata = data.Get("use_annotations_as_alias_metadata").(bool)
```
tomhjp, line 204: wrt to the lifecycle comment above, if the plugin gets restarted, it will wipe the value of this setting, because the `kubeAuthBackend` will get recreated and we aren't durably storing the provided `use_annotations_as_alias_metadata` setting anywhere. If we move this into `kubeConfig` to include it in the storage write just above, then it will be durably stored and survive plugin restarts.

integrationtest/integration_test.go
```diff
@@ -100,7 +100,38 @@ func annotateServiceAccount(t *testing.T, name string, annotations map[string]st
...
+		}
+	}
+
+	defer func() {
+		if t.Failed() {
+			cleanup()
+		}
+	}()
+
+	return cleanup
```
tomhjp, line 131: nit: I think you can use t.Cleanup to simplify a bit

```suggestion
	t.Cleanup(func() {
		_, err = client.Logical().Delete(fmt.Sprintf("/sys/policy/%s", name))
		if err != nil {
			t.Fatal(err)
		}
	})
```

integrationtest/integration_test.go
```diff
@@ -148,18 +185,42 @@ func setupKubernetesAuth(t *testing.T, boundServiceAccountName string, mountConf
...
-	return client, cleanup
+func setupKv1Mount(t *testing.T, client *api.Client, path string) func() {
+	_, err := client.Logical().Write(fmt.Sprintf("/sys/mounts/%s", path), map[string]interface{}{
+		"type": "kv",
+	})
+	if err != nil {
+		t.Fatalf("Expected to enable kv secrets engine but got: %v", err)
+	}
+
+	cleanup := func() {
```
tomhjp, line 202: I think this can probably be simplified with t.Cleanup too?

integrationtest/integration_test.go
```diff
@@ -148,18 +185,42 @@ func setupKubernetesAuth(t *testing.T, boundServiceAccountName string, mountConf
...
 
-	_, err = client.Logical().Write("auth/kubernetes/role/test-role", roleConfig)
+	_, err := client.Logical().Write("auth/kubernetes/role/test-role", roleConfig)
 	if err != nil {
 		t.Fatal(err)
 	}
+}
 
-	return client, cleanup
+func setupKv1Mount(t *testing.T, client *api.Client, path string) func() {
```
tomhjp, line 194: nit: Go casing conventions

```suggestion
func setupKVV1Mount(t *testing.T, client *api.Client, path string) func() {
```

integrationtest/integration_test.go
```diff
@@ -246,20 +315,47 @@ func TestFailWithBadTokenReviewerJwt(t *testing.T) {
...
+		t.Fatalf("Expected successful auth configuration GET but got: %v", err)
+	}
+
+	mountAccessor, ok := secret.Data["accessor"]
+	if !ok {
+		t.Fatal("Expected auth configuration GET response to have \"accessor\"")
+	}
+
+	const policyNameFoo = "alias-metadata-foo"
+	const kvPath = "root"
```
tomhjp, line 347: nit: root is a bit of a confusing path to use, because it's quite an overloaded term

```suggestion
	const kvPath = "kv-v1"
```

integrationtest/integration_test.go
```diff
@@ -268,63 +364,39 @@ func TestAuthAliasMetadataAssignment(t *testing.T) {
 		t.Fatalf("Expected successful login but got: %v", err)
 	}
 
-	// query the entity alias and match up its custom metadata
-	secret, err := client.Logical().List("identity/entity-alias/id")
+	// verify that the templated policy works by creating key value pairs at root/data/foo with the kubernetes auth token
+	kv1MountCleanup := setupKv1Mount(t, client, kvPath)
+	defer kv1MountCleanup()
```
tomhjp, line 369: nit: I would move this call up just before creating the policy for better readability, keeping the most closely related bits closer together.

integrationtest/integration_test.go
```diff
@@ -246,20 +315,47 @@ func TestFailWithBadTokenReviewerJwt(t *testing.T) {
...
+	mountAccessor, ok := secret.Data["accessor"]
+	if !ok {
+		t.Fatal("Expected auth configuration GET response to have \"accessor\"")
+	}
+
+	const policyNameFoo = "alias-metadata-foo"
+	const kvPath = "root"
+	createPolicy(t, policyNameFoo,
+		fmt.Sprintf(`path "%s/{{identity.entity.aliases.%s.metadata.key-1}}"
+	{ capabilities = [ "read", "update", "create" ] }`, kvPath, mountAccessor))
```
tomhjp, line 350: Maybe worth pulling the const policy string (including the %s directives) into its own `const` because the strange indenting pattern here makes it quite odd to read.

tomhjp (commented):
The test looks great! I left a few nits and style comments, but I think there is 1 bug that is worth addressing.

backend.go
thyton: Thank you for pointing this out!

tomhjp (approved):
LGTM! 🎉

awnumar:
👋🏽 Just wanted to point out, the `example` is incorrect: its annotations have a `auth-metadata.` prefix.

PR #247, opened 2024-05-20
Prepare for v0.19.0 release

kpcraig:
This is the Release/changelog PR for vault-plugin-auth-kubernetes.

Of note here are updates to go-jose 2.6.0 to 2.6.3 (a direct dependency), and go-jose 4.0.1 to 4.0.3 (an indirect dependency) to resolve GO-2024-2631.


fairclothjm (commented):
Looks like we need to bump go-jose otherwise we won't get past security scanner for the release

```
make dev && scan binary bin/vault-plugin-auth-kubernetes
CGO_ENABLED=0 go build -o bin/vault-plugin-auth-kubernetes cmd/vault-plugin-auth-kubernetes/main.go
go: downloading k8s.io/api v0.29.3
Scanned file:{path:"bin/vault-plugin-auth-kubernetes"} in 4.7s - found 1 result(s)
  » Go Modules Scanner
    ⚠︎ found OSV reported vulnerability GO-2024-2631 in github.com/go-jose/go-jose/v4@v4.0.1
        bin/vault-plugin-auth-kubernetes
```

CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## Unreleased
+## 0.19.0 (May 20, 2024)
```
Zlaticanin, line 1:
```suggestion
## Unreleased

## 0.19.0 (May 20, 2024)
```

I think we can leave the Unreleased title above and just add the version below. Similar to what was done [here](https://github.com/hashicorp/vault-plugin-auth-azure/pull/155/files)

kpcraig:
go-jose v4.0.2 seems to still upset the scanner, although https://pkg.go.dev/vuln/GO-2024-2631 seems to think it was resolved.

@fairclothjm says go-jose is releasing another patch update next week, so maybe things will be better then?

kpcraig:
It looks like v2.6.3 of go-jose avoids the need to try to update to v3 or v4, although it does require a rename of the import from `gopkg.in/square/go-jose` to `github.com/go-jose/go-jose` (`gopkg.in/go-jose/go-jose` is also valid, but if we do move to v4, this is more future-proof)

benashz (approved):
LGTM!

PR #248, opened 2024-05-21
Update jose v4

kpcraig:
This is a short PR to update the go-jose library to v4 to avoid possible issues with v2 support, copying https://github.com/hashicorp/vault-plugin-auth-gcp/pull/203

It's possible v2.6.3 also fixes this issue (see https://github.com/go-jose/go-jose/releases/tag/v2.6.3), but this feels more future-proof (and the solution was readily available).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,9 @@
 ## Unreleased
 
 ### Changes
-
+* Updated `github.com/go-jose/go-jose/v3` and `github/go-jose/go-jose/v2` to v4.0.2
 * Updated dependencies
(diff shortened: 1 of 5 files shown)


vinay-gopalan (approved):

kpcraig:
Sorry @benashz , i didn't see the review tag - if you want some more changes I can put them in before the release tag

benashz:
@vinay-gopalan  @kpcraig - yes, I am fairly certain these changes will break depending on the K8s API server config.

benashz (commented):
We need to revisit this change.

PR #251, opened 2024-06-14
Update go and k8s versions

tvoran:
Update to go 1.22.4 keeping with [vault@main](https://github.com/hashicorp/vault/blob/main/.go-version), and test with k8s versions 1.26-1.30. Also configure dependabot for hashicorp actions.


.github/dependabot.yaml
```diff
@@ -0,0 +1,21 @@
+---
```
benashz, line 1: 👍

benashz (approved):
LGTM!

tvoran:
Thanks!

PR #269, opened 2024-11-11
bump k8s version

hashiblaum:
bump k8s version, remove 1.26

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -27,13 +27,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        kind-k8s-version: [1.27.13, 1.28.9, 1.29.10, 1.30.6, 1.31.2]
+        kind-k8s-version: [1.27.16, 1.28.15, 1.29.10, 1.30.6, 1.31.2]
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - name: Create K8s Kind Cluster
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        kind-k8s-version: [1.26.15, 1.27.13, 1.28.9, 1.29.4, 1.30.0]
+        kind-k8s-version: [1.27.13, 1.28.9, 1.29.10, 1.30.6, 1.31.2]
```
tvoran, line 30: It looks like there are some slightly newer versions available now for [1.28](https://hub.docker.com/r/kindest/node/tags?name=1.28) and [1.27](https://hub.docker.com/r/kindest/node/tags?name=1.27):

```suggestion
        kind-k8s-version: [1.27.16, 1.28.15, 1.29.10, 1.30.6, 1.31.2]
```

tvoran (commented):
Might as well update the kind version in tests.yaml too:

https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/29b4907c32fae05c3c401a3c99e3e27545024ad6/.github/workflows/tests.yaml#L36

and the KIND_K8S_VERSION in the Makefile:
https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/29b4907c32fae05c3c401a3c99e3e27545024ad6/Makefile#L6

tvoran (approved):
Feel free to add a note to the Unreleased section of the changelog for the k8s versions.

PR #298, opened 2025-05-30
Update changelog and go version for v0.22.1

jaireddjawed:
Update changelog and go version for new release


CHANGELOG.md
```diff
@@ -1,15 +1,19 @@
 ## Unreleased
 
-### Build
-* Build with go 1.23.7
+
+## 0.22.1 (May 30, 2025)
+
+## Build
```
Zlaticanin, line 6: Small nit: But let's make sure that `### Build` is the same header size as `### Dependency updates`, as opposed to the version (`##`)

PR #300, opened 2025-06-02
Prevent authentication into Vault without an audience

jaireddjawed:
# Overview

Currently, Vault's Kubernetes authentication method does not require roles to specify an audience. As a result, an attacker could potentially craft a malicious JWT from a compromised or rogue Kubernetes cluster and use it to authenticate against Vault, especially if the associated cluster does not validate the `aud` claim.

This pull request introduces a change that requires the `audience` field to be specified when creating or updating a Kubernetes role. Additionally, authentication attempts will fail if they use a role that lacks a configured audience.

Since this is a breaking change, it will be included in Vault v1.21+. However, another [PR](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/301) will be backported into older Vault versions that logs a warning when a role does not have an audience.
 
### Impact on users

* When a role is created, it is now required to provide an audience.
* Existing roles will need to be updated to have an audience. Roles can't be updated without providing an audience first. Existing role data is intact and remains unaffected otherwise.
* An authentication attempt that uses a role that does not have an audience configured will fail with the following error: 
`"role does not have an audience defined, please update the role to include an audience"`.

### Manual Testing

Testing was similar to the testing done in [PR 301](https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/301). However, errors are expected instead of warnings.

* Set up Kind cluster and Vault dev server.
* Configured Kubernetes authentication and created role.
* Verified that configuring a role without an audience yields an error.

```
vault write auth/kubernetes/config \
    token_reviewer_jwt="$(cat vault-sa-token)" \
    kubernetes_host="$KUBE_HOST" \
    kubernetes_ca_cert=@ca.crt
Success! Data written to: auth/kubernetes/config
jaired.jawed@jaired k8s-demo % vault write auth/kubernetes/role/my-role \
    bound_service_account_names=my-app \
    bound_service_account_namespaces=default \
    policies=my-policy \
    ttl=24h
Error writing data to auth/kubernetes/role/my-role: Error making API request.

URL: PUT http://127.0.0.1:8200/v1/auth/kubernetes/role/my-role
Code: 400. Errors:

* "audience" can not be empty
```

### DIFF
--- a/path_login.go
+++ b/path_login.go
@@ -354,7 +354,7 @@ func (b *kubeAuthBackend) parseAndValidateJWT(ctx context.Context, client *http.
 	if strings.TrimSpace(role.Audience) == "" {
 		return nil, logical.CodedError(http.StatusBadRequest,
 			fmt.Sprintf("Role %s does not have an audience defined. Please update the role to include an audience", roleName))
-  }
+	}
 
 	expected.Audiences = []string{role.Audience}
 


path_role.go
```diff
@@ -334,9 +334,9 @@ func (b *kubeAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logical
 		return logical.ErrorResponse("can not mix %q with values", "*"), nil
 	}
 
-	// optional audience field
-	if audience, ok := data.GetOk("audience"); ok {
-		role.Audience = audience.(string)
+	role.Audience = data.Get("audience").(string)
+	if strings.TrimSpace(role.Audience) == "" {
```
jaireddjawed, line 338: This change breaks how the audience is processed during role creation or during a role update. I think we can put this in 1.20+ and probably make another PR for 1.19 and below that simply logs a warning if a kubernetes role does not have an audience.

tsaarni:
Hey, just checking if you’ve had a chance to consider my concern with this change in my earlier comment. Would love to hear your thoughts when you get a chance!

jaireddjawed:
@tsaarni I have gotten feedback from the team and we plan to do the following.

* Add a configurable setting on the Kubernetes role config to require audiences if a Vault admin prefers to do so, but it's default will be `false`.
* Simply log a warning when an audience is not defined so that admins are aware.

